PR #8, opened 2024-02-23
Add ok and alert to image rendering

OmegaRogue:
### Description
This PR implements support for rendering the `ok` and `alert` Action states in OpenDeck.

### Changes made

    `showOk` and `showAlert` Parameters added to `renderImage`

### Breaking changes

None

### DIFF
--- a/src/lib/rendererHelper.ts
+++ b/src/lib/rendererHelper.ts
@@ -59,7 +59,7 @@ export async function renderImage(actionContext: string, state: ActionState, sho
 	if (showOk) {
 		let okImage = document.createElement("img");
 		okImage.crossOrigin = "anonymous";
-		okImage.src = "ok.png";
+		okImage.src = "/ok.png";
 		await new Promise((resolve) => {
 			okImage.onload = resolve;
 		});
@@ -69,7 +69,7 @@ export async function renderImage(actionContext: string, state: ActionState, sho
 	if (showAlert) {
 		let alertImage = document.createElement("img");
 		alertImage.crossOrigin = "anonymous";
-		alertImage.src = "alert.png";
+		alertImage.src = "/alert.png";
 		await new Promise((resolve) => {
 			alertImage.onload = resolve;
 		});
--- a/src/lib/rendererHelper.ts
+++ b/src/lib/rendererHelper.ts
@@ -18,7 +18,7 @@ export function getImage(image: string | undefined, fallback: string): string {
 	return image;
 }
 
-export async function renderImage(actionContext: string, state: ActionState) {
+export async function renderImage(actionContext: string, state: ActionState, ok: boolean, alert: boolean) {
 	// Create canvas
 	let canvas = document.createElement("canvas");
 	canvas.width = 144;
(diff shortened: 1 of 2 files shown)


src/lib/rendererHelper.ts
```diff
@@ -18,7 +18,7 @@ export function getImage(image: string | undefined, fallback: string): string {
 	return image;
 }
 
-export async function renderImage(actionContext: string, state: ActionState) {
+export async function renderImage(actionContext: string, state: ActionState, ok: boolean, alert: boolean) {
```
ninjadev64, line 21: Please use more descriptive variable names, like `showOk` and `showAlert`

src/lib/rendererHelper.ts
```diff
@@ -35,6 +35,20 @@ export async function renderImage(actionContext: string, state: ActionState) {
 		image.onload = resolve;
 	});
 
+	let alertImage = document.createElement("img");
+	alertImage.crossOrigin = "anonymous";
+	alertImage.src = "http://localhost:5173/alert.png";
```
ninjadev64, line 40: http://localhost:5173 is only the Vite dev server and is not accessible in a release build or if a different port is being used

src/lib/rendererHelper.ts
```diff
@@ -35,6 +35,20 @@ export async function renderImage(actionContext: string, state: ActionState) {
...
+	await new Promise((resolve) => {
+		alertImage.onload = resolve;
+	});
+
+	let okImage = document.createElement("img");
+	okImage.crossOrigin = "anonymous";
+	okImage.src = "http://localhost:5173/ok.png";
```
ninjadev64, line 47: Same here

src/lib/rendererHelper.ts
```diff
@@ -55,7 +69,12 @@ export async function renderImage(actionContext: string, state: ActionState) {
 			context.fillRect(x - (width / 2), y + 2, width, 3);
 		}
 	}
-
+	if(alert) {
```
ninjadev64, line 72: Please add a space after the `if` keyword

src/lib/rendererHelper.ts
```diff
@@ -35,6 +35,20 @@ export async function renderImage(actionContext: string, state: ActionState) {
 		image.onload = resolve;
 	});
 
+	let alertImage = document.createElement("img");
```
ninjadev64, line 38: These images should only be loaded inside their respective if statements below to prevent unnecessary network usage and performance impact. Additionally, please use a consistent order; `ok` seems to come before `alert` in the existing codebase, the same way you've added them to the signature of this function.

ninjadev64 (changes requested):

ninjadev64:
Additionally please add a description to this pull request's body, see #7 for a rough template

src/lib/rendererHelper.ts
OmegaRogue: What else should i use then? I tried using `getImage`, but that lead to CORS errors

PR #29, opened 2024-11-14
Support installing plugins through deep linking

DesignCodeCreate:
Support for installing plugins from marketplace through deeplinking.


src-tauri/capabilities/main.json
ninjadev64, line 1: This file should be indented with tabs

src/components/PluginManager.svelte
```diff
@@ -121,6 +122,11 @@
 	(async () => plugins = await (await fetch("https://ninjadev64.github.io/openaction-plugins/catalogue.json")).json())();
 
 	let search: string = "";
+
+	onOpenUrl((urls: any) => {
+		let id = urls[0].split("installPlugin/")[1];
+		installPluginGitHub(id, plugins[id]);
+	});
```
ninjadev64, line 129: Please could you add checks for when the string does not contain installPlugin or the plugin ID is not present in the plugins object, and return doing nothing when this is the case

ninjadev64 (changes requested):

DesignCodeCreate:
Has been resolved

ninjadev64:
LGTM, merged :+1:

PR #93, opened 2025-06-19
PTT support - Allow KeyDown events to leave keys pressed by not dropping Enigo context

Dessix:
This PR alters the `key_down` and `key_up` functions in `input_simulation.rs` to not drop the Enigo context between invocations, reducing latency of invocations and allowing for Held (e.g. Push-To-Talk style)  key bindings in OpenDeck.
This also reduces mapping reloads by keeping `xmodmap` keys mapped between invocations, instead of triggering Enigo's cleanup with every interaction.

A mutex is added around the Enigo context to safeguard against concurrent access by multiple calls.

### DIFF
--- a/plugins/com.amansprojects.starterpack.sdPlugin/src/main.rs
+++ b/plugins/com.amansprojects.starterpack.sdPlugin/src/main.rs
@@ -46,7 +46,9 @@ impl openaction::ActionEventHandler for ActionEventHandler {
 	) -> EventHandlerResult {
 		match &event.action[..] {
 			"com.amansprojects.starterpack.runcommand" => run_command::down_up("down", event),
-			"com.amansprojects.starterpack.inputsimulation" => input_simulation::key_down(event),
+			"com.amansprojects.starterpack.inputsimulation" => {
+				input_simulation::down_up(event, "down").await
+			}
 			_ => Ok(()),
 		}
 	}
@@ -58,7 +60,9 @@ impl openaction::ActionEventHandler for ActionEventHandler {
 	) -> EventHandlerResult {
 		match &event.action[..] {
 			"com.amansprojects.starterpack.runcommand" => run_command::down_up("up", event),
-			"com.amansprojects.starterpack.inputsimulation" => input_simulation::key_up(event),
+			"com.amansprojects.starterpack.inputsimulation" => {
+				input_simulation::down_up(event, "up").await
+			}
 			"com.amansprojects.starterpack.switchprofile" => {
 				switch_profile::key_up(event, outbound).await
 			}
--- a/plugins/com.amansprojects.starterpack.sdPlugin/src/input_simulation.rs
+++ b/plugins/com.amansprojects.starterpack.sdPlugin/src/input_simulation.rs
@@ -1,33 +1,26 @@
 use openaction::*;
 
+use std::sync::LazyLock;
+
 use enigo::{
 	Enigo, Settings,
 	agent::{Agent, Token},
 };
+use tokio::sync::Mutex;
 
-pub fn key_down(event: KeyEvent) -> EventHandlerResult {
-	if let Some(value) = event.payload.settings.as_object().unwrap().get("down") {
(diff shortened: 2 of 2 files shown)


nekename:
Hey, thanks for the PR! I'll look into this soon. According to the enigo developer this might also help alleviate an issue with the libei support :)

nekename:
However, right off the bat I can say that you should use a tokio Mutex as this project deals with async code, and you should be able to drop the Arc by passing in a ref to the locked MutexGuard from GlobalEventHandler :)

Even better, use a static LazyLock\<Mutex>, then this can be isolated to input_simulation.rs.

nekename:
Also, please write commit messages using Conventional Commits and sign them with your signing key.

plugins/com.amansprojects.starterpack.sdPlugin/src/main.rs
```diff
@@ -119,7 +127,28 @@ async fn main() {
 	)
 	.unwrap();
 
-	if let Err(error) = init_plugin(GlobalEventHandler {}, ActionEventHandler {}).await {
+	let enigo = match Enigo::new(&Settings {
+		release_keys_when_dropped: true,
```
pentamassiv, line 131: You can just derive the default for the whole `Settings` struct. `release_keys_when_dropped: true` is the default value (see https://github.com/enigo-rs/enigo/blob/cf10baddde4b267b858c94ff08f697572d26a532/src/lib.rs#L481).

I only briefly looked at the code, but this is better usage of the `Enigo` struct. Recreating it each time leads to a number of issues as it keeps track of important state.

Dessix:
> However, right off the bat I can say that you should use a tokio Mutex as this project deals with async code, and you should be able to drop the Arc by passing in a ref to the locked MutexGuard from GlobalEventHandler :)

I'll do this- I was using the standard sync version as a bit of a just-in-case around what would've been a `Cell`, because I'm assuming no concurrency, but the overhead from a tokio lock is minimal enough that it's not a problem.

> Even better, use a static LazyLock<Mutex>, then this can be isolated to input_simulation.rs.

I'm not a huge fan of this idea because I wanted to keep state around in a way that wasn't necessarily globally bound, but I think a reasonable middle-ground would be to have this be an object with internally mutable state instead of a namespace with static methods. This lets us add the "reinitialize on failures" logic I [mentioned in another comment thread](https://github.com/nekename/OpenDeck/pull/93#discussion_r2157727831) without lifting much of it into the parent module. Would that be an acceptable compromise?

plugins/com.amansprojects.starterpack.sdPlugin/src/main.rs
pentamassiv: I never used enigo for programs that run a long time, so I am also not sure if there can be any problems that occur after hours or so. If they are recoverable, it surely makes sense to attempt to recreate it. There shouldn't be much that can go wrong though. If you do encounter problem like that, please let me know so I am aware of it.

As for my other comment, sure you can be explicit. I don't know the code base or the context of the code. I just thought I'd mention it.

nekename:
> I think a reasonable middle-ground would be to have this be an object with internally mutable state instead of a namespace with static methods

This sounds OK, but would be strangely different from the rest of the actions. Any reason you didn't want globally bound state?

Dessix:
Mostly that static globals feel ... Wrong, in Rust?

nekename:
Lol, I've done many so not for me

nekename:
```rust
static ENIGO: LazyLock<Mutex<Enigo>> = LazyLock::new(|| Mutex::new(Enigo::new(...)));
```
should suffice I think? Would greatly simplify the code

nekename:
@Dessix If you don't mind, I took it upon myself to rewrite this, as I'm going away soon and wanted to have this merged before I left. I think the way I have done it is reasonably clean, and I've made sure that it leaves the keys pressed now.

@pentamassiv You were right, this fixes the libei only working once issue! I'll comment with more details on the relevant enigo issue

Issue #94, opened 2025-07-03
Kimai plugin not working


atxbyea:
Hi and thanks for this great product.

I've been using some basic features for a while, but I recently discovered Kimai and want to use that.

However the plugin does not behave, and I cannot see anything in the opendeck logs either, any insight here would be very helpful

https://github.com/b263/stream-deck-time-tracker

Says successfully signed in, however, no dropdown of projects exist

![Image](https://github.com/user-attachments/assets/4023b6f9-c110-4557-a558-59e30458bf89)

![Image](https://github.com/user-attachments/assets/f25f44cf-ac5f-4279-9616-7ca2d34336c1)

nekename:
Please turn on Developer Mode, restart OpenDeck, try to use the plugin, then right click > inspect and check the console in the main OpenDeck window, as well as the window opened for the plugin.

Invisi:
Looking at the source code of Kimai[0], that's likely the same issue I just ran into with the Home Assistant plugin[1].
The regex for `image/svg+xml` in `rendererHelper.getImage` doesn't handle image data with `;`, or in this case `;charset=utf8`.

https://github.com/nekename/OpenDeck/blob/12dbce745fb750c4415a563fdea617ddf3f91f83/src/lib/rendererHelper.ts#L10-L14

I just fixed that locally by changing the code as follows:
```diff
-	const svgxmlre = /^data:image\/svg\+xml,(.+)/;
+	const svgxmlre = /^data:image\/svg\+xml(;[\w=]*)?,(.+)/;
	const base64re = /^data:image\/(apng|avif|gif|jpeg|png|svg\+xml|webp|bmp|x-icon|tiff);base64,([A-Za-z0-9+/]+={0,2})?/;
	if (svgxmlre.test(image)) {
-		image = "data:image/svg+xml;base64," + btoa(decodeURIComponent((svgxmlre.exec(image) as RegExpExecArray)[1].replace(/\;$/, "")));
+		image = "data:image/svg+xml;base64," + btoa(decodeURIComponent((svgxmlre.exec(image) as RegExpExecArray)[2].replace(/\;$/, "")));
	}
```

Can throw that in a PR real quick, unsure whether that might cause issues elsewhere but it looked fine in regex101
https://regex101.com/r/DprOEl/1


[0] https://github.com/b263/stream-deck-time-tracker/blob/c100c71ec43b2bcad2c4932c1a9f017e947a9de7/src/js/src/lib/tracker.ts#L124
[...]

PR #95, opened 2025-07-03
fix: make SVG image regex more lenient

Invisi:
See #94

This should capture all formats whilst ignoring `data` URLs that are already base64-encoded.

Samples to test with:
https://regex101.com/r/DprOEl/2
```
data:image/svg+xml;charset=utf8,<svg>
data:image/svg+xml;utf8,<svg>
data:image/svg+xml;,<svg>
data:image/svg+xml,<svg>
data:image/svg+xml;base64,SGVsbG8gd29ybGQ=
```

### DIFF
--- a/src/lib/rendererHelper.ts
+++ b/src/lib/rendererHelper.ts
@@ -7,7 +7,7 @@ export function getImage(image: string | undefined, fallback: string | undefined
 	if (!image) return fallback ? getImage(fallback, undefined) : "/alert.png";
 	if (image.startsWith("opendeck/")) return image.replace("opendeck", "");
 	if (!image.startsWith("data:")) return "http://localhost:57118/" + image;
-	const svgxmlre = /^data:image\/svg\+xml(?:;\w+=\w+)*,(.+)/;
+	const svgxmlre = /^data:image\/svg\+xml(?!.*?;base64.*?)(?:;[\w=]*)*,(.+)/;
 	const base64re = /^data:image\/(apng|avif|gif|jpeg|png|svg\+xml|webp|bmp|x-icon|tiff);base64,([A-Za-z0-9+/]+={0,2})?/;
 	if (svgxmlre.test(image)) {
 		image = "data:image/svg+xml;base64," + btoa(decodeURIComponent((svgxmlre.exec(image) as RegExpExecArray)[1].replace(/\;$/, "")));


nekename:
The data URL spec is as follows:

[3](https://www.rfc-editor.org/rfc/rfc2397#section-3). Syntax

       dataurl    := "data:" [ mediatype ] [ ";base64" ] "," data
       mediatype  := [ type "/" subtype ] *( ";" parameter )
       data       := *urlchar
       parameter  := attribute "=" value

   where "urlchar" is imported from [[RFC2396](https://www.rfc-editor.org/rfc/rfc2396)], and "type", "subtype",
   "attribute" and "value" are the corresponding tokens from [[RFC2045](https://www.rfc-editor.org/rfc/rfc2045)],
   represented using URL escaped encoding of [[RFC2396](https://www.rfc-editor.org/rfc/rfc2396)] as necessary.

   Attribute values in [[RFC2045](https://www.rfc-editor.org/rfc/rfc2045)] are allowed to be either represented as
   tokens or as quoted strings. However, within a "data" URL, the
   "quoted-string" representation would be awkward, since the quote mark
   is itself not a valid urlchar. For this reason, parameter values
   should use the URL Escaped encoding instead of quoted string if the
   parameter values contain any "tspecial".

   The ";base64" extension is distinguishable from a content-type
   parameter by the fact that it doesn't have a following "=" sign.

So some of the examples you gave that passed the regex were invalid (e.g. `data:image/svg+xml;utf8,<svg>`), and in addition, your regex didn't match the possibility of multiple parameters (e.g. `data:image/svg+xml;charset=utf8;foo=bar,<svg>`). I've updated the regex to this: https://regex101.com/r/DprOEl/3 which I think is compliant to the spec.

Please could you test my changes and let me know if this is good to merge?

Invisi:
No clue where I got `data:image/svg+xml;utf8,<svg>` from ~~and I can't find any plugin (on Github) using that syntax.~~
Some plugins may up ending using that, it's quite common:
https://github.com/search?q=%22data%3Aimage%2Fsvg%2Bxml%3Butf8%2C%3Csvg%22&type=code

e.g here's one of them
https://github.com/ViMaSter/streamdeck-tapthetile

Works on my machine, should be good to go as it matches the spec now. Thanks :)

nekename:
Oh interesting, but I think that if we find a plugin using that then the plugin should be fixed to be compliant, not the other way around. As long as the Home Assistant, Discord and Kimai plugins are working (these were the three affected) then it should be good to go - I'm going to wait until @Terrorwolf01 has tested my changes (he confirmed yours were working) and then I'll merge it!

nekename:
So Terrorwolf had issues with the Home Assistant plugin with my commit, the images look like this: `data:image/svg+xml;,<svg>`. So I essentially reverted to something closer to your original commit, the only change now being that it accepts multiple parameters whereas yours only accepted one.

Edit: final regex: https://regex101.com/r/DprOEl/4

PR #100, opened 2025-07-11
UX Updates

titusdmoore:
Note: These changes have been tested on both light and dark mode without a device connected. I've tested as best I can in areas that only show for device settings, but there may be some effects that I missed.

Change the following UX pain-points:
- Update hover state, icon, and sizing around actions details element.
- Update button hover effects on main screen.
- Update button hover effects on settings popup.
- Update icon hover effects in plugin list.
- Update popup close icon hover effect.
- Updated icons to use tailwind colors instead of hex colors.


nekename:
Closing because stale + a few of these things have been implemented separately since

PR #102, opened 2025-07-12
Fix left clicking the tray icon on Windows

MMonkeyKiller:


nekename:
Hey @MMonkeyKiller - I presume without this check, it shows the window on mouse down and then hides immediately on mouse up? 

I'll be happy to merge this, just please fix the blocked merge (your commit must have a verified signature).

MMonkeyKiller:
Hello! I've just fixed the unsigned commit thing, it should be good now.
Yes, without this fix, left clicking on the tray icon (at least on Windows) would show and hide the window immediately. Now it only toggles the visibility only when the left click is on key_down state

nekename:
Thank you so much!

PR #103, opened 2025-07-13
Hide webview window by default

MMonkeyKiller:
By doing this, I solve a bug in windows when the program starts up. For a second you can see the window flashing before it hides


nekename:
I'll test this out on Linux soon and will be happy to merge it, thanks for the PR!

nekename:
Could u lmk if it's still all fine on Windows after ce1551a (should be bc code stayed the same for Windows but just anyways)

PR #105, opened 2025-07-16
Prevent plugins from running after they are unregistered

MMonkeyKiller:
This fixes a bug in Windows on which unregistering a plugin would throw an `Access Denied` because of the plugin still running on the background while trying to delete it.
It also unregisters all the plugins when the program exits, preventing them from running in the background after the application is closed, causing problems when trying to reopen the websocket connection.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -264,6 +264,7 @@ Enjoy!"#,
 
 	app.run(|app, event| {
 		if let tauri::RunEvent::Exit = event {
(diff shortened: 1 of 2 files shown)


nekename:
Thanks for another PR! I could have sworn that I had already implemented this: I knew about this bug but didn't implement this solution to fix it because I was certain I already had. Apparently not, though. I didn't mean to push the version bump commit yesterday so I've just done your commit again so I can merge it before the version bump - I'll try and make sure GitHub attributes it to you.

PR #106, opened 2025-07-17
Implement the `systemDidWakeUp` event

MMonkeyKiller:
This PR introduces the integration of a new dependency ([`psp`](https://github.com/pewsheen/psp)), and implements functionality for handling system wake-up events. Specifically, it introduces the [`systemDidWakeUp`](https://docs.elgato.com/streamdeck/sdk/guides/system/#system-wake) event, which is broadcast to compatible plugins, enhancing support for Stream Deck plugins.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -150,7 +150,17 @@ dependencies = [
  "serde_repr",
  "tokio",
  "url",
- "zbus",
+ "zbus 5.7.1",
+]
+
+[[package]]
+name = "async-broadcast"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


nekename:
I'm sorry I haven't replied to you on this. This will certainly be included in 2.6.0

nekename:
Hey @MMonkeyKiller - having looked at the psp dependency, I've had second thoughts about this. It hasn't been updated in 2 years, isn't published to crates.io, and uses very outdated versions of dependencies, which introduces duplicate dependencies into OpenDeck and increases size and build times, and also might be a security issue. For example, it uses zbus 3 which is on 5.9 now, or event-listener 2.5 when OpenDeck currently has 5.4.

If you can find another crate, or get this one updated to the newest dependencies, I'll have another look

MMonkeyKiller:
Hello! I understand, I will look for a new crate that accomplishes the same thing as this one. If I happen to find one I'll update this pr

PR #109, opened 2025-07-27
Show property inspector scrollbars only when needed

MMonkeyKiller:
This fixes a bug where scrollbars would be visible even when no proper inspector was shown

Before:

After:


nekename:
Thanks for another great contribution!

PR #111, opened 2025-08-01
Stop key canvases from flickering

MMonkeyKiller:
This pull request fixes a visual bug where canvases would clear the displayed image before loading the new image, causing it to flicker for a fraction of a second.

Before:
opendeck_xbleutr1yd

After:
opendeck_ARBRnib9LR


nekename:
🙏 never thought the fix for this would be this simple

PR #114, opened 2025-08-04
Fix bugs when dragging an action

MMonkeyKiller:
This pull request fixes some bugs related to dragging actions out of the action list
 - Makes the entire action draggable, not just the action image
 - Prevents the action category name from being selected, which eliminates visual bugs when dragging
 
 Before:
opendeck_VoDj2snrNh

 After:
opendeck_WEt195y9Lm

### DIFF
--- a/src/components/ActionList.svelte
+++ b/src/components/ActionList.svelte
@@ -12,18 +12,21 @@
 	reload();
 </script>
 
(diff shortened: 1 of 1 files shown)


nekename:
I moved the select-none up a level because I could still select the text by starting the drag on an empty region. Thanks for another nice enhancement

PR #122, opened 2025-08-16
Implement the `applicationDidLaunch` and `applicationDidTerminate` events

MMonkeyKiller:
This pull request adds a new dependency ([sysinfo](https://crates.io/crates/sysinfo)), and implements event handling for process launch and termination, which may be required for certain plugins.

It introduces two new events, `applicationDidLaunch` and `applicationDidTerminate`, which can be used by plugins to respond to the launch and termination of specific applications. For more details, refer to the [Stream Deck documentation](https://docs.elgato.com/streamdeck/sdk/guides/app-monitoring/).

Additionally, a new `ApplicationsToMonitor` property has been added to the plugin manifest, allowing plugins to specify which applications they want to monitor.

This functionality has been tested on Windows only.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2864,6 +2864,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ntapi"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8a3895c6391c39d7fe7ebc444a87eb2991b2a0bc718fdabd071eec617fc68e4"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "num-conv"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


nekename:
Thanks for another nice contribution. I changed variable names to make it more understandable, changed an `if-continue` into an `if-else`, prevented a deadlock that would have occurred if stop_monitoring_all was called (it in turn called stop_monitoring which locked the same RwLock), changed the information reported by sysinfo to `nothing` for performance, and made the API of application_monitor.rs more friendly for plugins/mod.rs.

Please review my code, then test it, and let me know if it's OK! Also, just curious, which plugin are you implementing these for? I haven't ever found a plugin that actually needs them, so just wanted to know

Another thing, this probably won't work on Mac, because the plugins expect bundle IDs and not process names. That's the reason I held off on implementing this because I had trauma from doing [this](https://github.com/nekename/OpenDeck/blob/5d4b3b002d72df9f8452ef89b3c8571b4a3f2ef6/src/script/plugins.ts#L301) in OpenDeck V1. But I'll merge this because it's been done now so might as well have an impl for Windows (and Linux if a Linux plugin ever needs it)

MMonkeyKiller:
Hello! Everything works fine, I've just tested it on my PC and still works. Thanks for making the code more readable!

Well, originally I was trying to make the original OBS Studio plugin compatible with OpenDeck, because I was having issues with the one created by BarRaider. I was almost finishing this when I realized I could just have used Multi OBS Controller 😅.

I didn't know you had already implemented this functionality in an older version. I'm sure this implementation will need some tweaks to make it compatible with Mac and Linux (and possibly Wine), but honestly I have little experience working with them.

nekename:
It's compatible with native Linux plugins from my testing, but there's only 4 plugins made natively for Linux right now. 

Does Elgato's OBS plugin work now?

Thanks for the contribution, as always

MMonkeyKiller:
Yes, the Elgato's OBS plugin works now. Thank you for merging it!

PR #123, opened 2025-08-18
fix: drop application_plugins before sleep

MMonkeyKiller:
This PR fixes a bug that could sometimes happen when exiting the app. The `stop_monitoring` function could block the whole shutdown logic freezing the application.


nekename:
Good catch, thanks, I didn't want to add a new commit so I amended the existing one from #122

PR #124, opened 2025-08-18
Display category icons in ActionList

MMonkeyKiller:
This PR changes the way categories are handled, allowing to display the respective category icons in the UI. 
- It introduces a new `CategoryInfo` structure that contains both the category's icon and its respective actions
- The frontend now supports optional category icons, displayed next to the category name

Before:
opendeck_zAbkqqd05N

After:
opendeck_YLnLDpDSLR


Terrorwolf01:
I think this is a good idea, but as in your Screenshot, it could be easily mistaken for another Action. I think we should something to be able to see it better whats a category or an Action except the small arrow.

nekename:
I made the icons much smaller, improved the margins, and made it use the plugin's icon if there's no category icon and all the actions in the category are from the same plugin.

Screenshot_20250819_164021
Screenshot_20250819_164108
Screenshot_20250819_164130

PR #125, opened 2025-08-19
fix(ui): only show actions meant to be visible

MMonkeyKiller:
This PR simply makes use of the `VisibleInActionsList` property, and uses it to filter the displayed actions shown in the actionlist

Before:
imagen

After:
imagen

### DIFF
--- a/src/components/ActionList.svelte
+++ b/src/components/ActionList.svelte
@@ -29,7 +29,7 @@
 				{/if}
 				<span class="ml-1">{name}</span>
 			</summary>
-			{#each actions as action}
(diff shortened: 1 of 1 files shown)


nekename:
VisibleInActionsList is intentionally ignored because the only reason an action would be hidden is if it was configured in a profile included with the plugin, and OpenDeck doesn't support those profiles (yet), so I list them in the action list so people can at least put the profiles together themselves.

If you would like to add support for the profile format used by plugins then feel free to do that (not an easy task though), but until that's implemented I'm going to keep ignoring VisibleInActionsList

Thanks though

PR #128, opened 2025-08-21
fix(ui): adjust width of the actionlist container

MMonkeyKiller:
This PR fixes a visual bug when searching for an action/category that doesn't exist.
(cherry picked from commit 86b4d77d7a2c487695adf7fdd11f2f1a0c1ba031)

Before:
opendeck_0Kbu0suJ3x

After:
opendeck_pYCGdR4iis

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -42,7 +42,7 @@
 	{/if}
 </div>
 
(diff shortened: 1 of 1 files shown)


nekename:
Thanks, I lumped this in with another related change I was going to make anyway

PR #129, opened 2025-08-21
fix(ui): prevent dragging only the action icon

MMonkeyKiller:
This PR fixes a simple visual bug on which dragging from the action icon would only show the icon being dragged instead of the whole draggable action.

Before:
opendeck_OV1H3pqx2x

After:
opendeck_VItX7xISjX


nekename:
Thanks

PR #131, opened 2025-08-22
fix: prevent socket handle inheritance on Windows

MMonkeyKiller:
This PR addresses an issue on Windows where forcefully killing OpenDeck would leave the socket in a `LISTENING` state, preventing OpenDeck from binding to the same port on restart. This happened because, by default, child processes inherit handles in Windows, causing the sockets to stay active even after the main process was terminated. The fix ensures that this problem no longer occurs by preventing socket inheritance.

This bug does not occur on Unix because `TcpListener` on Unix already sets the `SO_REUSEADDR` option, which allows for rebinding the same port even if a process is terminated.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -3216,7 +3216,7 @@ dependencies = [
  "tokio",
  "tokio-tungstenite",
  "urlencoding",
- "windows 0.61.3",
+ "windows-sys 0.60.2",
  "zip",
 ]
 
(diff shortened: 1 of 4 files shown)


nekename:
Windows is a strange operating system...

Thanks

nekename:
@MMonkeyKiller does this make #105 no longer necessary?

MMonkeyKiller:
Not really, because this PR handles the worst case scenario (like if the program crashes/is killed with task manager for example). On the other hand, the other PR handles a graceful shutdown of the plugins, making sure that all the plugins are killed before exiting the program

PR #135, opened 2025-08-24
Add keyboard and screen reader support for streamdeck keys, key context menus and the action list

zersiax:
Let me preface this by saying I'm no ninja coder, I'm a developer mostly well-versed in accessibility. There may very well be better ways to do some of the things I do in this PR and if you'd rather I just maintain my own fork that's fine, I just wanted to make sure people had a way to use these devices if they use a keyboard or a screenreader, as right now they don't.

This pull request:

- Allows for arrow-key-navigation within the key grid as well as the action list
- Pressing enter on an action in the action list stores the action, pressing enter on a key associates the action with that key. Pressing enter again on the same key opens the editor.
- Pressing f10, or the applications key on a WIndows/Linux? machine opens the context menu, suspends key code propagation to the keys and routes focus to the menu so it can be used with the arrow keys and enter key.
- Delete on a key clears it, and ctrl+c and ctrl+v should work for copying actions from one key to another.
- I think I made sure to visually indicate what key has keyboard focus visually but I can't really verify that (blind myself, heh)
The PR SHOULD pass all linting and formatter checks, I ran them several times throughout my edits.
I'm going to allow edits by maintainers in case things aren't up to the standards of the project :-)

### DIFF
--- a/src/components/ActionList.svelte
+++ b/src/components/ActionList.svelte
@@ -4,6 +4,9 @@
 	import { localisations } from "$lib/settings";
 
 	import { invoke } from "@tauri-apps/api/core";
+	import { createEventDispatcher } from "svelte";
+
+	const dispatch = createEventDispatcher();
 
 	let categories: { [name: string]: { icon?: string; actions: Action[] } } = {};
 	let plugins: any[] = [];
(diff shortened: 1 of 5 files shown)


nekename:
Closing as stale - I think I had said I'd be happy to merge this if it was split into chunks to be easier to review

Issue #141, opened 2025-09-15
Please expand input fields for Run Command


AtmanActive:
Hi,

It is all working fine, no bugs here, but, the input fields for 'Run Command' built-in action are too short and not responsive, which leads to frustration when working with them.

Most of the time, the command I want to run is long and the current input field, restricted to short length is forcing me to select-scroll left and right many times in order to figure out what is the value stored inside. It is really hard to use.

The culprit is the div container `.sdpi-item` which is constrained to` max-width: 344px;`.
If that could be made responsive, so it expands when I expand the main OpenDeck window, that would be awesome.

Thanks.

AtmanActive:
In my local "Inspect Element" hacking tests, just removing the CSS declaration `max-width: 344px;` from `.sdpi-item` (`spdi.css` line 340) resolves the issue nicely.

Of course, there's no point in me hacking my runtime, it needs to be solved in the source.

Thanks.

Terrorwolf01:
Hi,
Feel free to submit a PR since in Cases like this the exact answer is already known.
We then can easier test it in different scenarios.
Another thing I personally use for longer commands is a .sh script where I put them in. Then I also have syntax highlighting available.

AtmanActive:
> Feel free to submit a PR

I can't even find `spdi.css` in the source.

AtmanActive:
Oops, spoke too soon.

Made [the PR](https://github.com/nekename/OpenDeck/pull/142).

Thanks.

PR #142, opened 2025-09-15
Update sdpi.css to remove max-width for input fields

AtmanActive:
Fix for issue #141


nekename:
Looks good. I might try to get rid of this file completely at some point, make the starter pack plugin's property inspector blend in with the rest of the OpenDeck UI.

PR #180, opened 2025-11-29
feat: add drag-and-drop functionality for image uploads in InstanceEditor

Alex-Jr:
This PR provides an easy way to edit action icons by adding drag-and-drop functionality.

https://github.com/user-attachments/assets/49f72fdf-635b-45b8-99c3-58fa84b6f53f

### DIFF
--- a/src/components/InstanceEditor.svelte
+++ b/src/components/InstanceEditor.svelte
@@ -15,25 +15,23 @@
 	let fileInput: HTMLInputElement;
 	let colourInput: HTMLInputElement;
 
-	let isDragging = false;
-
 	function handleDrop(event: DragEvent) {
 		event.preventDefault();
-		isDragging = false;
 
 		const file = event.dataTransfer?.files?.[0];
-		if (file && file.type.startsWith("image/")) {
-			const reader = new FileReader();
-			reader.onload = async () => {
-				let result = reader.result?.toString();
-				if (result) {
-					let resized = await resizeImage(result);
-					if (resized) instance.states[state].image = resized;
-					else instance.states[state].image = result;
-				}
-			};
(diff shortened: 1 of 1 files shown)


src/components/InstanceEditor.svelte
```diff
@@ -15,6 +15,27 @@
 	let fileInput: HTMLInputElement;
 	let colourInput: HTMLInputElement;
 
+	let isDragging = false;
```
nekename: I'm not sure I understand the utility of this variable. You don't seem to access its value anywhere. Maybe it (this line and all the lines that reference it) can just be removed entirely?

src/components/InstanceEditor.svelte
```diff
@@ -43,6 +64,16 @@
 	<div class="flex flex-row">
 		<div class="flex flex-col justify-center items-center">
 			<button
+				on:drop={handleDrop}
+				on:dragstart={() => {
+					isDragging = true;
```
nekename: Same here, but also, the `ondragstart` event is fired when the element its attached to starts to be dragged, i.e. when someone starts dragging this button element with the current image, which isn't possible because it isn't dragging. I think you can just remove this block.

src/components/InstanceEditor.svelte
```diff
@@ -43,6 +64,16 @@
...
+				}}
+				on:dragover={(event) => {
+					event.preventDefault();
+				}}
+				on:dragleave={() => {
+					isDragging = false;
```
nekename: Same here

nekename (changes requested):
Thank you so much for your contribution!

I've left a few comments.

PR #185, opened 2025-12-01
Add VS Code Debugging Support and Refactor Action Move/Copy Logic

Alex-Jr:
- Add launch.json and tasks.json
Provides debugging support in VS Code, following the recommended steps from the [Tauri Debugging Guide](https://v2.tauri.app/develop/debug/vscode/)

image

- Update the move_instance command
Rewrites the function to allow swapping actions between two keys.

[Screencast_20251201_075910.webm](https://github.com/user-attachments/assets/160e1ed6-f08d-4680-a044-f855da494ac1)

- Create copy_instance command
Used during copy and paste operations to improve separation of concerns, simplifying move_instance command

This is my first time writing Rust at a more serious level, so any feedback or suggestions for improvement are greatly appreciated.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,31 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"type": "lldb",
+			"request": "launch",
+			"name": "Debug (debug)",
+			"cargo": {
+				"args": [
+					"build",
+					"--manifest-path=./src-tauri/Cargo.toml",
+					"--no-default-features"
+				]
+			},
+			"preLaunchTask": "frontend:dev"
+		},
+		{
+			"type": "lldb",
+			"request": "launch",
+			"name": "Debug (release)",
+			"cargo": {
+				"args": [
+					"build",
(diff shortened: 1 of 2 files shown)


nekename:
Hi @Alex-Jr: before I can take a look at this, the changes need to be in separate PRs. You can make each change on a separate branch based on the main branch to do this.

Alex-Jr:
Okay @nekename, I will start with the commit adding the new copy command, otherwise the move instance changes will break copy and paste

nekename:
To be honest, I would prefer you implement the action swapping without the separation of copy, as that results in more code duplication and also more changes to review and potentially introduce issues.

So just one PR for action swapping using the existing move_instance, and another for VSCode debugging. Thanks :)

Alex-Jr:
I originally tried to keep everything inside a single function, but that made move_instance more complicated. It had to handle three different behaviors: moving, swapping, and copying, each with different effects on both profile and filesystem. It also felt odd for a function named move to not be moving anything, only copying.

PR #187, opened 2025-12-01
feat: implement copy_instance command and update DeviceView to use it…

Alex-Jr:
- Create copy_instance command
Used during copy and paste operations to improve separation of concerns, simplifying move_instance command

### DIFF
--- a/src-tauri/src/events/frontend/instances.rs
+++ b/src-tauri/src/events/frontend/instances.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use super::Error;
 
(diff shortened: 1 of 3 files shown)


nekename:
I'm sorry, but I still stand by what I said earlier. I prefer minimal changes to the code and minimal duplication because I don't have the capacity to maintain so much of it, so if copying is working right now, I'd like to keep it that way.

If you want to implement the swapping without this change that will gladly be accepted. Also, I've merged your PR for the debugging configs, that's really useful! I've just tried it out and it seems like it will be very helpful for me actually

PR #196, opened 2025-12-15
Implement `didReceiveDeepLink` event

MMonkeyKiller:
This pull request introduces a new event: `didReceiveDeepLink`, following [ElGato's DeepLink implementation](https://docs.elgato.com/streamdeck/sdk/guides/deep-linking/). 
It supports `opendeck://` and `streamdeck://` schemes. Tested only on Windows.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -282,14 +282,14 @@ If you have already donated, thank you so much for your support!"#,
 				&& let Ok(url) = reqwest::Url::parse(&args[pos])
 				&& let Some(mut path) = url.path_segments()
 			{
-				if !url.query_pairs().any(|(k, v)| k == url.scheme() && v == "hidden") {
-					let _ = show_window(app);
-				}
-
 				if url.host_str() == Some("plugins")
 					&& path.next() == Some("message")
 					&& let Some(plugin_id) = path.next()
 				{
+					if !url.query_pairs().any(|(k, v)| k == url.scheme() && v == "hidden") {
+						let _ = show_window(app);
+					}
+
 					let plugin_id = if url.scheme() == "streamdeck" { format!("{plugin_id}.sdPlugin") } else { plugin_id.to_owned() };
 					tauri::async_runtime::spawn(async move {
 						if let Err(error) = events::outbound::deep_link::did_receive_deep_link(&plugin_id, args[pos].clone()).await {


nekename:
Hey @MMonkeyKiller and thanks for another great contribution. Are there any plugins that I could test this with, or that it benefits/makes supported?

Also, I'm going to keep the `opendeck` scheme for internal use (like `install-plugin`) to prevent any conflicts with the paths that Elgato might introduce and to allow us to do what we want with it. I'll add an `openaction` scheme - that way, we can match Elgato 1:1 and also keep it a bit more vendor-agnostic at the same time.

nekename:
There, that's my changes. I've changed .to_string to .to_owned() (https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/5) and handled the error without .unwrap(). Let me know if that looks alright and what plugins I can test this with :)

MMonkeyKiller:
Hello! Yeah, I wasn't sure if opendeck:// was meant for this. I was trying to use [Simple Macro](https://github.com/Hajsori/Simple-Macro_Stream-Deck) when I realized deep links weren't a thing yet. 
Keep in mind that if you want to install this plugin to test this thing out you'll have to rename the folder inside `opendeck/plugins` and remove the `.sdPlugin` suffix, otherwise the Minecraft mod is going to send the plugin message to an invalid plugin uuid

nekename:
> going to send the plugin message to an invalid plugin uuid

Ah, of course - we need some way of handling that in the code then, because no deep links will work otherwise - OpenDeck unfortunately includes the .sdPlugin suffix in internal ID tracking.

Could you take a look at that?

MMonkeyKiller:
Yeah sure. I could check if the scheme is `streamdeck://` and then append `.sdPlugin` to the received plugin id. Is it ok?

nekename:
Yeah, I guess that works

nekename (approved):

nekename:
Thanks a lot!

I'll test the changes out soon and merge. What are the chances that I play Minecraft for the first time in a year on a server where it's my first time using SVC and you come out with this?

MMonkeyKiller:
Must be some kind of sign 😅. I hope you find this PR useful

nekename:
Pushed a small fix, but it looks like it's working with my testing plugin. Thanks!

nekename:
More coverage of the API is always nice. Hopefully we don't lose too many plugins because of Elgato's DRM.

PR #200, opened 2025-12-21
Built in icon gallery with `.streamDeckIconPack` support

NecroKote:
# What
Allow users select from internal icon gallery. The gallery consists of icons from Installed icon packs. For now only native StreamDeck packs are supported (`.streamDeckIconPack`)

Hold <kbd>Alt</kbd> (<kbd>Option</kbd> on mac) while clicking on icon to open build in icon selector. 
Selector supports search by icon name and icon pack name.

https://github.com/user-attachments/assets/be2f9fe4-6e3d-4a2c-a054-cf4432bc464a

# How
- new tauri module `iconpacks` with custom `icon://` protocol
- managed `IconPackManager` as state
  - `fst`-based search. Hopefully allowing thousands of installed icons to be searchable in an instant. Probably a total mess and definitely an overkill
- custom tauri frontend commands that call `IconPackManager`
- new set of frontend components. Too many to mention

I've barely tested this on my mac, and it somehow works 🤷 
Totally needs to be polished a bit before merging.

Let me know what you think @nekename.

### DIFF
--- a/src/components/IconPacks/IconPackManager.svelte
+++ b/src/components/IconPacks/IconPackManager.svelte
@@ -59,21 +59,23 @@
 
 <IconPickerModal bind:open={showIconPreview} />
 
-<Modal bind:open={showPopup}>
+<Modal bind:open={showPopup} class="w-10/12">
 	<h2 slot="header" class="font-semibold text-xl dark:text-neutral-300">
 		Manage Icon Packs
 	</h2>
 
 	<div slot="children" class="flex flex-col gap-4 mt-4 mb-4">
-		<div class="flex flex-row items-center">
-			<button
-				class="flex flex-row gap-1 px-2 py-0.5 items-center text-sm text-neutral-700 dark:text-neutral-300 bg-neutral-100 dark:bg-neutral-700 border dark:border-neutral-600 rounded-lg outline-hidden"
-				on:click={() => { showIconPreview = true; }}
-			>
-				<MagnifyingGlass />
-				<span>Explore installed icons</span>
-			</button>
-		</div>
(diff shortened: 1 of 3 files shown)


nekename:
Wow, that's a lot of stuff! Thanks!

My first concern is that the Elgato Marketplace now has DRM "protection" on all asset files, which if it doesn't apply to icon packs yet, it definitely will soon.

That aside, it's a lot of new code, so I will wait until you have tested it thoroughly and cleaned up any of the "mess" you mentioned, because otherwise I will be spending a lot of time reviewing it again and again whenever you make a change!

I would also want to let you know that I am crazy about making sure everything is the same everywhere in a codebase - that means, for example, using the same Tailwind classes where possible, or the same import groupings, so just keep that in mind :)

I think that the alt+click UX is kinda clanky. Maybe change the "solid colour" button to a row of three icon buttons: solid colour, icon picker, and reset to default. That way we can avoid the alt+click and remove the right-click to reset.

I'll take a skim over the code now but will leave my detailed review to when you think it's ready :)

Thanks again

nekename:
Ok, I've just skimmed over the Rust code very very briefly.

The formatting is all over the place with a mix of tabs and spaces. Please follow the checklist in the contributing section of the README to format and lint all of your code.



Secondly, the use statements / imports seem poorly organised: please organise them like the rest of the codebase (super, crate, std, other). You're also using .into() to convert &str to String whereas the rest of the codebase uses .to_owned()

I've only looked for 2 mins on my phone tho and not at all at the frontend

nekename:
Stale

PR #208, opened 2026-01-02
Stream Deck Neo Touch Points

mcbridet:
image

Adds **touch point** support. 

Touch points are display-less buttons found on the **Stream Deck Neo**, and they are implemented so that they work more or less the same as regular buttons; anything that can be assigned to a button can also be assigned to a touch point.

The touch points can be assigned a colour, this PR uses what is set on the "Solid colour" option, or as derived from the average colour of the slot (if there's an image set).

image

---

Basic LCD support has also been added for future use. At the moment it just displays "OpenDeck" logo (generated in the project from text, not an image).

image

Better than nothing I guess.

### DIFF
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -19,7 +19,7 @@ fn extract_average_colour(img: &image::DynamicImage) -> (u8, u8, u8) {
 	let (r_sum, g_sum, b_sum) = img
 		.pixels()
 		.fold((0u64, 0u64, 0u64), |(r, g, b), (_, _, pixel)| (r + pixel[0] as u64, g + pixel[1] as u64, b + pixel[2] as u64));
-	let count = (img.width() * img.height()).min(1) as u64;
+	let count = (img.width() * img.height()).max(1) as u64;
 	((r_sum / count) as u8, (g_sum / count) as u8, (b_sum / count) as u8)
 }
 


src-tauri/src/elgato.rs
```diff
@@ -63,6 +89,7 @@ pub async fn set_brightness(brightness: u8) {
 
 pub async fn reset_devices() {
 	for (_id, device) in ELGATO_DEVICES.read().await.iter() {
+		clear_all_touchpoints(device).await;
```
nekename: this is the thing im talking about in my comment

nekename (commented):
Hey @mcbridet 

First, I removed the branding thing because tbh, it only suggests to the user that something can be done with the screen when in reality it can't be, and adds a non-configurable "OpenDeck" to their setup which they probably don't want. Besides, that code would just be removed if we get the LCD support in anyway, so it's not like it's doing any useful prep work.

Then, I made the line rendered on the canvas into an element, because that way it matches the border colour properly in both light and dark modes, and is also less code and more efficient.

Finally, I made extract_average_colour more concise, and cleanup up some placement and naming of code.

I've just got one question: is the clearing of the touchpoints when resetting the device necessary? I would have thought that the device.reset() call would have done that automatically. Please test if that's actually necessary, and if it's not, then remove it (this reset code is used when quitting the app, btw).

Other than that, it all looks good to me now, so please test my changes in case I broke anything, and thanks for adding this! Quite a few people have asked about this so this will make them happy :)

mcbridet:
@nekename Thanks for reviewing. I'll re-visit LCD stuff another day, did you have a deck with a strip?

Regarding the changes to extract_average_colour, the changes to not appear to work for me.

image

I would expect the touch points to be green on the left and either white-blue or blue on the right.

Cheers.

mcbridet:
image
Eggcellent. :) Looks good to me.!

nekename:
yippee :tada:

Issue #213, opened 2026-01-07
Additional action appearance customisation features


maxtron95:
Hello, I've recently tried out this app. After a few days of using it, I wanted to request some features as to customize it some more for my purposes.

- [ ]  Linux system icon theme picker (Linux-only)
- [ ]  Image adjustment options (Move/Scale/Rotate/Colorize)
- [x]  Allow for a BG color underneath image/icon
- [x]  Load custom font (From system or file)
- [x]  Text outline options (Color/Toggle outline on or off)

Terrorwolf01:
For the Font, you can already use the System Fonts if I am not mistaken.

maxtron95:
> For the Font, you can already use the System Fonts if I am not mistaken.

Maybe... It's that when I go to the font dropdown in OpenDeck, I only get to choose from these specific fonts. Though I have a lot more fonts installed on my system in the font browser. (Using Fedora KDE)

<img width="254" height="266" alt="Image" src="https://github.com/user-attachments/assets/3b312c63-1332-4e5a-a064-a6f28f41741d" />

nekename:
For the font, you just type the name if it's not in the dropdown. But I agree, it would be better if they were in the dropdown.

nekename:
#217 solves the font list one, thanks to @clxxiii

nekename:
Forgot to update here, but #216 by @clxxiii has been merged and will be released with OpenDeck v2.9.0.

nekename:
I have removed the plugin installation progress bar from the list because it doesn't fit in with the other features.

nekename:
No. 3 done by @ArcticTechy in #233

nekename:
Number 1 and 2 seem pretty far away, ngl, potentially even out of the scope of the project... will think about whether I close this issue here now or not.

nekename:
I think I have decided that number 1 and 2 are out of the scope, so I'll close this as completed thanks to the great work of our contributors to implement 3, 4 and 5.

PR #215, opened 2026-01-07
feat: load font list from system installed fonts

clxxiii:
This PR works on fixing the QOL issues raised in #213 

For the font listing thing, I had to add a dependency called [`font-loader`](https://crates.io/crates/font-loader). On MacOS and Windows it seems to work just fine, on unix it needs a system dependency called "libfontconfig", which is already required by "wine" and "webkit2gtk4.1", so it shouldn't be an issue.

### DIFF
--- a/src/components/InstanceEditor.svelte
+++ b/src/components/InstanceEditor.svelte
@@ -174,17 +174,17 @@
 					class="w-full px-1 dark:text-neutral-300 bg-neutral-200 dark:bg-neutral-600 rounded-md outline-hidden"
 				/>
 				<datalist id="families">
+					<option value="Liberation Sans">Liberation Sans</option>
+					<option value="Archivo Black">Archivo Black</option>
+					<option value="Comic Neue">Comic Neue</option>
+					<option value="Courier Prime">Courier Prime</option>
+					<option value="Tinos">Tinos</option>
+					<option value="Anton">Anton</option>
+					<option value="Liberation Serif">Liberation Serif</option>
+					<option value="Open Sans">Open Sans</option>
+					<option value="Fira Sans">Fira Sans</option>
+					<option disabled>──────────</option>
 					{#each fonts as font}
-						<option value="Liberation Sans">Liberation Sans</option>
-						<option value="Archivo Black">Archivo Black</option>
-						<option value="Comic Neue">Comic Neue</option>
-						<option value="Courier Prime">Courier Prime</option>
-						<option value="Tinos">Tinos</option>
-						<option value="Anton">Anton</option>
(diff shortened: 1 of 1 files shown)


src-tauri/src/main.rs
```diff
@@ -66,6 +66,7 @@ async fn main() {
 		.invoke_handler(tauri::generate_handler![
 			frontend::restart,
 			frontend::get_devices,
+			frontend::get_fonts,
```
nekename: Again, this should go after set_application_profiles

src-tauri/Cargo.toml
```diff
@@ -51,6 +51,7 @@ sysinfo = { version = "0.37", default-features = false, features = ["system"] }
 semver = "1.0"
 path-slash = "0.2"
 fix-path-env = { git = "https://github.com/tauri-apps/fix-path-env-rs", rev = "c4c45d503ea115a839aae718d02f79e7c7f0f673" }
+font-loader = "0.11.0"
```
nekename: Could this be 0.11 instead of 0.11.0 to match the other dependencies here

src/components/InstanceEditor.svelte
```diff
@@ -12,6 +13,13 @@
 	let bold: boolean;
 	let italic: boolean;
 
+  let fonts: string[] = [];
+
+  onMount(async () => {
+    fonts = await invoke("get_fonts");
+    console.log(fonts);
+  });
+
```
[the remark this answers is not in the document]
nekename: Also the console.log should be removed

src/components/InstanceEditor.svelte
```diff
@@ -167,15 +175,9 @@
...
-					<option value="Comic Neue">Comic Neue</option>
-					<option value="Courier Prime">Courier Prime</option>
-					<option value="Tinos">Tinos</option>
-					<option value="Anton">Anton</option>
-					<option value="Liberation Serif">Liberation Serif</option>
-					<option value="Open Sans">Open Sans</option>
-					<option value="Fira Sans">Fira Sans</option>
+          {#each fonts as font}
+            <option value={font}>{font}</option>
+          {/each}
```
nekename: Will this also remove the options for the built in fonts that were there before? I think you should restore the options that were there before and add a `<option disabled>──────────</option>` in between to separate the built in fonts from the system fonts

nekename (changes requested):
Overall, looks good, I've requested some very minor changes, also could each change be its own PR (so open new PRs for the other checklist items), that way I can squash any changes needed per improvement into one commit for that improvement

PR #217, opened 2026-01-08
feat: load font list from system installed fonts

clxxiii:
Messed around with the branches on my fork and it accidentally closed the [old PR](https://github.com/nekename/OpenDeck/pull/215) so here's a new one.
> This PR works on fixing the QOL issues raised in https://github.com/nekename/OpenDeck/issues/213
> 
> For the font listing thing, I had to add a dependency called [font-loader](https://crates.io/crates/font-loader). On MacOS and Windows it seems to work just fine, on unix it needs a system dependency called "libfontconfig", which is already required by "wine" and "webkit2gtk4.1", so it shouldn't be an issue.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -711,6 +711,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cmake"
+version = "0.1.57"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


nekename:
Thanks! It works nicely. Gonna have to rebase the commits to get them signed because GitHub is unhappy.

nekename:
Ok, squashed them manually which signed it with my signature. Thanks again :)

PR #219, opened 2026-01-11
fix: add bounds check for state index

nerodoescode:
prevents a panic @ states.rs "index out of bounds: the len is 1 but the index is 1".


nekename:
You missed a few spots :)

PR #222, opened 2026-01-13
Add link to OpenDeck Portable edition for windows

AtmanActive:
Add link to OpenDeck Portable edition for windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,8 @@ Special thanks go to the developers of [Tauri](https://github.com/tauri-apps/tau
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


nekename:
Hi,
I appreciate the effort but I'm unlikely to include this in any official fashion because I'm too busy to have to think about more things that might be broken when I change things.
If it's unofficial then it's fine if stuff breaks, but I have to consider it if it becomes more official.
Also, your use of a Markdown link to point "GitHub Releases" to your repo without specifying it's a third-party repo is a little sketchy.
Thanks and apologies

PR #233, opened 2026-01-28
Key bg color

ArcticTechy:
I added Background rendering to the button and made the preview in the inspectorEditor a canvas to allow for showing the background colour and render other stuff like text in the future "hint"


ArcticTechy:
forgot a picture
image
image
image

PR #234, opened 2026-01-28
Render text in instance editor preview

ArcticTechy:
Added code to the getInstanceEditor function to render text in the preview


ArcticTechy:
image
image

PR #236, opened 2026-01-31
Feat: Render Text in key state preview editor

ArcticTechy:
This will render what we do in our text editor inside InstanceEditor


nekename:
This can be closed now that #237 has been merged

PR #250, opened 2026-02-18
Update 40-streamdeck.rules

Bwitzig:
Adding new u.dev rule for Ajazz usb device ordered from Temu, confirmed working on Cachyos.


Bwitzig:
Updated udev rule once again as I didn't realize I needed it in two places :)

nekename:
No? You must submit your changes to the udev rules file of whichever plugin you are using

PR #257, opened 2026-02-28
Add keyboard focus and accessible labels to device keys

zersiax:
## Summary
- Make canvas-based keys focusable by adding `tabindex="0"` and `role="button"`. I will progressively enhance this later to make it a proper grid, trying to not make the PRs too large and this is enough to work from for now.
- Add `aria-label` to each key with its grid position (e.g. "Key A1"), action name, and custom text when visible. This is needed because contents of canvases are generally screenreader-inaccessible.
- Fix `keyup` handler to only trigger on Enter/Space instead of any key press, to enable for arrow-based navigation later.

Verified working on my local, both with keyboard and screen reader.

### DIFF
--- a/src/components/Key.svelte
+++ b/src/components/Key.svelte
@@ -156,7 +156,7 @@
 		height={size}
 		draggable={slot != null}
 		tabindex="0"
-		role="button"
+		role="gridcell"
 		aria-label={accessibleLabel}
 		on:dragstart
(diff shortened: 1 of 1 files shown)


nekename:
Ignore Copilot it's being silly

nekename:
Thanks @zersiax!

PR #260, opened 2026-03-01
Fix property inspector message handlers referencing undefined event

tbrodeen:
## Summary
- The injected message handlers in the property inspector iframe destructured the `MessageEvent` parameter to just `{ data }`, but then referenced `event` for `stopImmediatePropagation()`. This caused a `ReferenceError`, preventing `connectElgatoStreamDeckSocket` from being called and leaving property inspectors stuck in a loading state.
- Changed the handlers to receive the full `event` parameter and extract `data` from it.

## Steps to reproduce
1. Install any plugin that uses `sdpi-components.js` for its property inspector (e.g. NeewerLite)
2. Add an action to the deck and click on it to open the property inspector
3. The property inspector stays stuck on "Loading..."

## Root cause
In `webserver.rs`, the two `window.addEventListener("message", ...)` handlers both destructure `({ data })` but then call `event.stopImmediatePropagation()` — since `event` is not in scope, this throws a `ReferenceError` before `connectElgatoStreamDeckSocket` can be called.

## Fix
Receive the full `event` parameter and extract `data` via `const data = event.data`.

### DIFF
--- a/src-tauri/src/plugins/webserver.rs
+++ b/src-tauri/src/plugins/webserver.rs
@@ -80,7 +80,8 @@ pub async fn init_webserver(prefix: PathBuf) {
 					const opendeck_window_open = window.open;
 					const opendeck_iframe_container = document.getElementById("opendeck_iframe_container");
 
-					window.addEventListener("message", ({ data }) => {
+					window.addEventListener("message", (event) => {
+						const data = event.data;
 						if (data.event == "connect") {
 							event.stopImmediatePropagation();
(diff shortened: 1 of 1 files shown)


nekename:
Thanks for the catch. I presume that this works normally because it references window.event (https://developer.mozilla.org/en-US/docs/Web/API/Window/event) - any idea why it doesn't work in your case specifically? Good to change this anyway because we shouldn't be using that property

PR #264, opened 2026-03-11
feat: Add configurable idle sleep for decks

FazalAAli:
## Summary
- add a configurable idle sleep timeout for devices in app settings
- put individual decks to sleep by dropping brightness to 0 after inactivity and wake them on button or encoder input
- persist the new setting in the Rust and TypeScript settings models and expose it in the settings UI

## Testing
- cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features
- deno check
- deno task check
- deno lint
- manual test: set sleep timeout to 1 minute, confirm settings.json persisted it, verify deck slept and woke on input

### DIFF
--- a/src-tauri/src/device_sleep.rs
+++ b/src-tauri/src/device_sleep.rs
@@ -30,7 +30,7 @@ pub fn update_timeout_minutes(minutes: u16) {
 	}
 }
 
-pub async fn register_device(device: &str) -> Result<(), anyhow::Error> {
+pub async fn note_activity(device: &str) -> Result<(), anyhow::Error> {
 	LAST_ACTIVITY.insert(device.to_owned(), Instant::now());
 	wake_device(device).await
 }
@@ -38,11 +38,6 @@ pub async fn register_device(device: &str) -> Result<(), anyhow::Error> {
 pub fn deregister_device(device: &str) {
 	LAST_ACTIVITY.remove(device);
 	SLEEPING_DEVICES.remove(device);
-}
-
-pub async fn note_activity(device: &str) -> Result<(), anyhow::Error> {
-	LAST_ACTIVITY.insert(device.to_owned(), Instant::now());
-	wake_device(device).await
 }
 
 async fn sleep_idle_devices() -> Result<(), anyhow::Error> {
--- a/src-tauri/src/events/inbound/devices.rs
+++ b/src-tauri/src/events/inbound/devices.rs
@@ -21,7 +21,7 @@ pub async fn register_device(uuid: &str, mut event: PayloadEvent<crate::shared::
 		event.payload.plugin = uuid.to_owned();
 		let _ = crate::events::outbound::devices::device_did_connect(&event.payload.id, (&event.payload).into()).await;
 		DEVICES.insert(event.payload.id.clone(), event.payload.clone());
-		let _ = crate::device_sleep::register_device(&event.payload.id).await;
+		let _ = crate::device_sleep::note_activity(&event.payload.id).await;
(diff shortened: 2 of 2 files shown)


FazalAAli:
Dear maintainer, 

I also have a `shell.nix`[ file ](https://nixos.wiki/wiki/Development_environment_with_nix-shell) that I created while working on this. Do you have any use for it? I can push up a separate PR for it.

TLDR: A shell.nix will allow contributors to use nix and create a working environment for this project in a reproducible manner. Meaning, all contributors will get the same package versions and set up to build your project.

When using the shell.nix, the build process (even on a fresh machine) simply becomes:
```
1. nix-shell shell.nix
2. deno task tauri dev
  ```

File: src-tauri/src/device_sleep.rs [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+	SLEEPING_DEVICES.remove(device);
+}
+
+pub async fn note_activity(device: &str) -> Result<(), anyhow::Error> {
+	LAST_ACTIVITY.insert(device.to_owned(), Instant::now());
+	wake_device(device).await
+}
+
+pub async fn input_started(device: &str) -> Result<(), anyhow::Error> {
+	let mut count = ACTIVE_INPUTS.entry(device.to_owned()).or_insert(0);
```
nekename: Is ACTIVE_INPUTS really necessary? Who is going to press a button and reset the timer and hold it all the way until the timer runs out - i think this is unnecessary added complexity

src-tauri/src/elgato.rs
```diff
@@ -93,6 +93,13 @@ pub async fn set_brightness(brightness: u8) {
 	}
 }
 
+pub async fn set_device_brightness(id: &str, brightness: u8) {
+	if let Some(device) = ELGATO_DEVICES.read().await.get(id) {
+		let _ = device.set_brightness(brightness.clamp(0, 100)).await;
+		let _ = device.flush().await;
+	}
+}
```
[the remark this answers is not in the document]
FazalAAli: Done, only one brightness function in elgato.rs now that takes in id and brigthness

device.rs now calls that function when iterating over devices. The new method is preserved and used by devices_sleep to put individual devices to sleep

src-tauri/src/events/outbound/devices.rs
```diff
@@ -91,22 +91,30 @@ struct SetBrightnessEvent {
 	brightness: u8,
 }
 
+//Set brightness for all devices, used by frontend/settings.rs to set global brightness
```
nekename: Comment badly formatted, run cargo fmt

src-tauri/src/elgato.rs
```diff
@@ -146,6 +146,20 @@ async fn init(device: AsyncStreamDeck, device_id: String) {
...
+				| DeviceStateUpdate::EncoderUp(_)
+				| DeviceStateUpdate::EncoderTwist(_, _) => {
+					if let Err(error) = crate::device_sleep::note_activity(&device_id).await {
+						log::warn!("Failed to wake sleeping device {device_id}: {error}");
+					}
+				}
+				_ => {}
+			}
 			match match update {
 				DeviceStateUpdate::ButtonDown(key) => keypad::key_down(&device_id, key).await,
```
nekename: Could you make these call inbound/devices.rs functions which you've added note_activity calls to, instead of calling outbound/keypad.rs (or encoder) functions directly? Then we don't need note_activity in elgato.rs

nekename:
Just rebased on main

nekename:
Excellent work @FazalAAli :)

PR #267, opened 2026-03-11
feat: trigger virtual button press on key double-click

FazalAAli:
## Summary
- add a frontend command that emits a virtual key/dial press for a slot context
- wire double-click on keys in the UI to trigger the virtual press
- register the new Tauri command in the app invoke handler

## Testing
- Built, and tested the feature

Feature in action
P.S. there is an animation that shows when you double click. It's just not showing up in the video for some reason

https://github.com/user-attachments/assets/36063946-ceca-457e-9aea-187b2b88de89


nekename:
@FazalAAli wondering if this should also use inbound/devices.rs's key_down etc functions so that they get the sleep PR's note_activity calls? I think it makes sense to wake the device on a virtual press

nekename:
Thanks @FazalAAli again!

PR #268, opened 2026-03-11
feat: Add custom image scale controls

FazalAAli:
## Summary
- add per-state custom image scale to action state data
- add compact + / - controls in the instance editor to adjust icon scale
- render scaled custom images centered in both preview and device output

## Validation
- cargo fmt
- deno check src/lib/rendererHelper.ts src/lib/ActionState.ts
- deno task check
-built and ran to test the feature

## Screenshots
Default:
image

Scaled up:
image

With default icon (Note how the text is unbothered):
image


nekename:
Notes to self:
- variable rename
- bypass

nekename:
Thanks @FazalAAli!

Issue #38, opened 2024-12-17
Statement regarding device support


ninjadev64:
There have been numerous requests for support for various pieces of hardware from various manufacturers over the past few months.

I would like to make it clear that it is not in my interests to spend time implementing interfaces to these devices, especially in the case of those that have comparatively few owners, and even if it were, it is not feasible for me to work on support for a device that I do not own as there is no way to verify that anything works, which results in an arduous, time-consuming back-and-forth with device owners.

For those devices that already have an implementation in some programming language, I have gone out of my way to make it incredibly easy to have these devices supported in OpenDeck by way of a plugin written in said programming language, even if it is not Rust, and am open to providing advice to any owner of said devices with programming skills to do so. For those that do not, I am afraid you will have to have reverse engineering skills and time or wait for someone else who has both.

I am now a maintainer of the upstream [elgato-streamdeck](https://github.com/streamduck-org/elgato-streamdeck) library which provides an interface to Elgato devices and those with similar firmware, notably from MiraBox and subsidiaries (including Ajazz). I will be reviewing and accepting PRs, but will not be making any of my own contributions.

glf999:
AKP153 written into StreamDock 293s firmware cannot be recognized

akp153 ==> 293s
VID 5548 PID 6670 [0300]

mattwarner1:
I'm so sorry to ask what might be a silly question. I see how to edit this library, but how do I compile it into the Open Deck code when I clone it?  I don't see instructions/steps in the documents, but if anyone could help step me through it, I'd really appreciate tips.

ninjadev64:
@mattwarner1 

You just need to replace the specification of the dependency in OpenDeck's Cargo.toml file.

So change it from `elgato-streamdeck = { version = "0.9", ... }` to `elgato-streamdeck = { path = "/path/to/cloned/library", ... }`

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

mattwarner1:
Thanks, @ninjadev64 , for the tips. I'll give it a shot when I'm a little fresher in the morning. :-)

> [@mattwarner1](https://github.com/mattwarner1)
> 
> You just need to replace the specification of the dependency in OpenDeck's Cargo.toml file.
> 
> So change it from `elgato-streamdeck = { version = "0.9", ... }` to `elgato-streamdeck = { path = "/path/to/cloned/library", ... }`
> 
[...]

PR #270, opened 2026-03-16
fix(macos): prevent WKWebView from suspending plugin js

gtfunes:
## Summary

On macOS, WKWebView windows created with `.visible(false)` have their JavaScript execution suspended by the OS after ~7 seconds as a power-saving measure. This causes all WebView-based plugins (like worldtime clock) to freeze. They stop sending WebSocket messages (SetTitle, SetImage, etc) and become completely unresponsive.

Native binary plugins are unaffected since they run as separate processes.

### Fix

A macOS-only background task that evaluates `void(0)` on each hidden WebView plugin window every 3 seconds, preventing macOS from suspending their JavaScript context. This is gated behind `#[cfg(target_os = "macos")]` so it has zero impact on Linux and Windows.

---

**Preflight checklist**

---

### DIFF
--- a/src-tauri/src/plugins/mod.rs
+++ b/src-tauri/src/plugins/mod.rs
@@ -447,15 +447,13 @@ pub fn initialise_plugins() {
 	#[cfg(target_os = "macos")]
 	tokio::spawn(async {
 		use tauri::Manager;
+		let app = APP_HANDLE.get().unwrap();
 		loop {
 			tokio::time::sleep(std::time::Duration::from_secs(3)).await;
 			let instances = INSTANCES.lock().await;
-			for (uuid, instance) in instances.iter() {
-				if matches!(instance, PluginInstance::Webview) {
-					let label = uuid.replace('.', "_");
-					if let Some(window) = APP_HANDLE.get().unwrap().get_webview_window(&label) {
-						let _ = window.eval("void(0)");
-					}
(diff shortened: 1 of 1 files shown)


nekename:
Looks great! Merging is blocked due to the lack of verified signatures. Wonder why I haven't seen this on my Mac, though

gtfunes:
@nekename oh, thanks for the heads up on the verified commits, I just fixed that.

I honestly noticed because I was trying to use mostly the HomeAssistant plugin and noticed it wasn't working at first. Then I noticed there was a pattern and it would work for a few seconds before stopping, then noticed other plugins like World Time would stop working too. Do JS dependant plugins work for you on mac?

gtfunes:
@nekename forgot to add: I tested this on two macbook pros (M1 & M4) both on Tahoe 26.3.1 and experienced the same issue. Not sure what macOS version you are using? Maybe it's something that has changed recently or related to power saving settings? Both macs were plugged in while testing, but could be something to keep in mind.

nekename:
Potentially a newer macOS issue, then, because my MacBook's on Sequoia - I had to do some strange stuff to the OS to make it allow me to set it up (got it from eBay, potentially stolen 😬) so it hasn't shown me any updates in a while

gtfunes:
@nekename hehe, could be then that it is something coming from the newer macOS versions. I believe it should be a harmless change then, even if running on an older macOS version, it shouldn't break plugins in any way.

nekename:
Thanks @gtfunes! :tada:

PR #276, opened 2026-03-22
feat: add keyboard grid navigation to the device view

zersiax:
## Summary

- Implements arrow key navigation between keys in the device view following the [WAI-ARIA grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/)
- Uses roving tabindex so the grid is a single Tab stop — arrow keys move between cells, Tab exits the grid entirely, and the last focused cell is remembered when tabbing back in
- Adds `role="grid"`, `role="rowgroup"`, and `role="row"` to the container structure so screen readers can convey the grid layout

## Keyboard interaction

| Key | Behaviour |
|---|---|
| Arrow keys | Move focus to adjacent cell (no wrapping at edges) |
| Home / End | First / last cell in current row |
| Tab | Exit grid to next focusable element |
| Shift+Tab | Exit grid to previous focusable element |
| Enter / Space | Select the focused key (unchanged) |

Column is clamped when moving to a row with fewer items (e.g. from the keypad row to the encoder row).

## Implementation notes

- Key.svelte receives a new `tabIndex` prop (default `-1`); DeviceView passes `0` only to the currently focused cell
- The keydown handler uses the capture phase on the grid container because Key.svelte uses `stopPropagation` on its own keydown/keyup handlers — capture intercepts arrow keys before they reach the cell, while Enter/Space pass through normally
- A `focusin` listener syncs the focused row/column state when a cell gains focus through click or Tab, keeping the roving tabindex in sync

### DIFF
--- a/src/components/Key.svelte
+++ b/src/components/Key.svelte
@@ -62,7 +62,7 @@
 		}
 	}
 
-	function focus() {
+	function onfocus() {
 		$openContextMenu = null;
 		if (!slot) return;
 		if (slot.action.uuid != "opendeck.multiaction" && slot.action.uuid != "opendeck.toggleaction") {
@@ -183,7 +183,7 @@
 		on:keyup|stopPropagation={(e) => {
 			if (e.key === " ") select(e);
 		}}
-		on:focus={focus}
+		on:focus={onfocus}
 		on:contextmenu={contextMenu}
 	/>
 	{#if isTouchPoint && !slot}


zersiax:
CoPilot feedback does not warrant changes. Comment 1 is valid, but I'm pretty sure all of those usages don't actually come from a point where those keys need to be focusable (active==false, context=null), unless I'm missing something. @nekename 
As for comment 2, it's somewhat inconsistent but not against the ARIA spec, fine to leave as is.
So this is good to merge whenever :)

PR #281, opened 2026-03-26
Fix/brightness knob flicker

mchampanis:
**Preflight checklist**

---

fix: reduce display flicker during brightness changes using encoder knob

On the Treslin N3 using the display brightness widget on an encoder knob, and then using that knob would cause display flickering (small flashes of black in between brightness changes). Using the UI slider rapidly with the mouse did the same.

I believe the following is a valid fix that is hardware agnostic.

  ## Summary
  - Debounce settings store subscriber to avoid flooding `set_settings` on rapid changes
  - Debounce brightness slider and encoder knob inputs in SettingsView
  - Extract `$settings.rotation` in Key.svelte so brightness changes don't trigger a full re-render of every
  button

  ## Test plan
  - [x] Turn brightness encoder knob rapidly -- no flicker, device brightness updates smoothly
  - [x] Drag brightness slider in settings -- no flicker or display going black
  - [x] Knob changes reflected in settings slider position
  - [x] Slider shows correct value on app restart
  - [x] Changing display rotation still re-renders keys correctly

### DIFF
--- a/src/components/Key.svelte
+++ b/src/components/Key.svelte
@@ -16,6 +16,8 @@
 	import { invoke } from "@tauri-apps/api/core";
 	import { listen } from "@tauri-apps/api/event";
 
+	$: rotation = $settings?.rotation;
+
 	export let context: Context | null;
 	export let label: string = "";
 	export let tabindex: number = 0;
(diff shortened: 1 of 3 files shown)


mchampanis:
I had an unrelated commit that I was testing in this branch by mistake - windows mutex stuff. Have fixed it now and made this new PR.

I'll test out the Key.svelte change by itself from your other comment: https://github.com/nekename/OpenDeck/pull/280#issuecomment-4136784953

nekename:
Yep, as I said, only the changes to Key.svelte *should* be necessary I believe

nekename:
@mchampanis did you get a chance to check if that's only what's necessary? v2.11.0 will be coming out very soon so just in case you want to get this in before that

nekename:
Let me know if you want this reopened

PR #288, opened 2026-03-29
fix: support symlinked plugin roots

JohnCMcDonough:
**Preflight checklist**
- [ ] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.

---
  ## Summary
  OpenDeck currently handles symlinked plugin roots inconsistently. Plugin processes can still start, but asset loading and some frontend-facing plugin discovery paths break once the plugin directory under the config folder is a symlink to a location outside that config tree.

 This change makes symlink handling consistent across plugin initialization and frontend discovery, and updates the plugin webserver to allow assets from resolved plugin roots while keeping the existing path guard in place.

  - allow plugin assets to load correctly when plugin directories in the config folder are symlinks
  - reuse a shared plugin entry path resolver across plugin initialization, frontend plugin discovery, localisation loading, and outbound plugin iteration
  - keep the webserver path guard in place while permitting files under resolved plugin roots

  ## Testing
  - cargo fmt --manifest-path src-tauri/Cargo.toml
  - cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features
  - deno check --unstable-raw-imports
  - deno task check
  - deno lint
  - launched `deno task tauri dev` and verified symlinked plugins load and register correctly

> Disclaimer: ***This PR was drafted by AI. I'm not a Rust Developer. I do not claim to fully understand the changes it's making.***

### DIFF
--- a/src-tauri/src/events/frontend/mod.rs
+++ b/src-tauri/src/events/frontend/mod.rs
@@ -77,10 +77,7 @@ pub async fn get_localisations(locale: &str) -> Result<HashMap<String, serde_jso
 	};
 
 	while let Ok(Some(entry)) = entries.next_entry().await {
-		let path = match entry.metadata().await.unwrap().is_symlink() {
-			true => tokio::fs::read_link(entry.path()).await.unwrap(),
-			false => entry.path(),
-		};
+		let path = crate::shared::plugin_entry_path(&entry.path()).unwrap_or_else(|_| entry.path());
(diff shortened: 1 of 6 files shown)


nekename:
This shouldn't be necessary for symlinked plugin paths to work - all you should need is to have developer mode enabled in settings. It's intentionally disallowed due to security concerns when developer mode isn't enabled. 

If it still isn't working when developer mode is enabled then please let me know

PR #290, opened 2026-03-31
feat: make the multi/toggle action child list keyboard accessible

zersiax:
## Summary
- Add keyboard navigation (Arrow Up/Down, Home, End) to the multi-action and toggle action child list using roving tabindex
- Focus automatically moves to the child list when the view opens, and focusing a child action loads its property inspector
- The drop zone accepts Ctrl+V to paste a copied action as a keyboard alternative to drag-and-drop
- Remove buttons are out of the tab order (Delete key on the row handles removal); accessible labels added to close and remove buttons
- Key.svelte gets a configurable `role` prop (defaults to `gridcell`) so it can render as `presentation` inside the list

### DIFF
--- a/src/components/ParentActionView.svelte
+++ b/src/components/ParentActionView.svelte
@@ -169,7 +169,7 @@
 		}}
 		role="listitem"
 		tabindex={children.length == 0 ? 0 : -1}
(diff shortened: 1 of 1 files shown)


nekename:
:tada:

PR #296, opened 2026-04-01
feat: Add  visual device preview rotation setting (portrait-friendly) and improve window sizing

flagadajones:
**Preflight checklist**
- [X ] I understand that if this pull request is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [ X] I have ensured that I have run the appropriate formatter on my changes and that my code produces no linter violations.
- [X ] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.
- [X ] I will keep "Allow edits from maintainers" enabled for this pull request.

---

This PR introduces a new user setting that rotates the device preview in the editor by 90-degree steps, independently from image rotation.

The goal is to make portrait layouts easier to visualize while keeping existing image rendering behavior unchanged.

### What Changed
- Added a new persisted setting: device preview rotation (0, 90, 180, 270).
- Added a new control in Settings for this value.
- Applied visual rotation to the full device preview layout so users can inspect configurations in portrait orientation.
- Kept image rotation logic separate (no behavior change for image rendering rotation).
- Fixed rotation direction so visual preview rotation matches the expected direction relative to image rotation.
- Updated window size calculations to use the effective visual dimensions after rotation, with larger margins so large devices are displayed more comfortably.

### Why
- Users need a clearer way to preview device setups in portrait mode.
- The previous sizing behavior could feel too small for larger layouts.
- Separating image rotation and preview rotation avoids side effects and keeps behavior intuitive.

### Validation
- Frontend build completed successfully.
- Rust backend check completed successfully.
- No component-level errors were reported in the modified UI files.

### Notes
- Existing settings remain backward-compatible through default values for the new field.
- This change is visual/editor-focused and does not alter hardware output behavior for key images.

### DIFF
--- a/src/components/ParentActionView.svelte
+++ b/src/components/ParentActionView.svelte
@@ -7,6 +7,7 @@
 	import Key from "./Key.svelte";
 
 	import { copiedItem, inspectedInstance, inspectedParentAction } from "$lib/propertyInspector";
+	import { settings } from "$lib/settings";
 
 	import { invoke } from "@tauri-apps/api/core";
 	import { onMount, tick } from "svelte";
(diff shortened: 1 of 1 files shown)


nekename:
Thank you for the contribution, but I'm really not sure that this feature is needed enough to justify the complexity that it adds to the code and the number of changes I would have to review. This would significantly increase maintenance burden for a purely cosmetic feature.

Thank you for submitting the other PR for GIF/WebP support, though, and I'll be sure to take a look at that for v2.12 as it's been requested often!

flagadajones:
I understand. I made this pull request because I use my Stream Deck in a vertical position, and it's simpler to configure it (position and rotation of images) so that I have the same rendering in Opendeck.

nekename:
That looks like a cool setup! Unfortunately yeah, too many code changes; if the changes were more simple I would have been happy to add the feature.

PR #302, opened 2026-04-05
fix: skip profile save for data URI images in setImage

### LINKED ISSUE #249: High write volume when action images update frequently
Hi,

I noticed that OpenDeck performs excessive and frequent write operations to the disk whenever a button's content changes (e.g., dynamic updates like CPU usage or system monitors). Using `iotop` and `strace`, I confirmed that the application writes multiple SVG files and a full JSON profile to the disk for every single state change, rather than caching them in memory.

## Reproduction Steps

1. **Environment:** Run OpenDeck on Linux with a plugin that generates frequent dynamic content (e.g., the [System Information Plugin](https://github.com/OpenActionPlugins/system) or my little [Streamdeck Matrix Plugin](https://github.com/Serveny/streamdeck-matrix) which can create some updates per second.
2. **Action:** Observe a button that updates at least once per second.
3. **Monitor Disk I/O:** Run the following command to trace write syscalls:
```bash
sudo strace -y -f -p $(pidof opendeck) -e trace=write,pwrite64 -s 100

```

4. **Monitor Throughput:** Use `iotop` to check accumulated disk writes.

## Observed Behavior

In a test window of only a few seconds, `iotop` reported **over 50 MB of accumulated writes**.

The `strace` log reveals that several `.svg` files and a `.json.temp` profile (~28KB per write) are being written to the config directory repeatedly for every update:

```text
/* High frequency writes to SVG assets and profile templates */
[pid 140629] write(83<.../images/sd-AL15I1C01181/Default/Keypad.9.0/0.svg>, "%3Csvg%20width%3D%2272%22...", 306) = 306
[pid 140629] write(83<.../images/sd-AL15I1C01181/Default/Keypad.11.0/0.svg>, "%3Csvg%20width%3D%2272%22...", 306) = 306
[pid 140629] write(83<.../profiles/sd-AL15I1C01181/Default.json.temp>, "{\n  \"keys\": [\n    {\n      \"action\": ...", 28527) = 28527

```

**Location of writes:** `~/.var/app/me.amankhanna.opendeck/config/opendeck/`

## Potential Impact

* **SSD Longevity:** Constant writing of 28KB+ files multiple times per second unnecessarily consumes **TBW (Total Bytes Written)**, potentially shortening the lifespan of NVMe/SSD drives.
* **Performance Bottlenecks:** Synchronous disk I/O on the main thread can lead to UI micro-stutters or input latency.

## Suggested Fixes

1. **In-Memory Asset Handling:** Temporary button icons (SVGs) should be kept in RAM or written to a volatile filesystem (like `/tmp` or `shm`) instead of the physical configuration directory.
2. **Atomic Updates:** Only write the JSON profile if the actual configuration structure has changed, not just the visual state of a dynamic button.

## System Information

* **OS:** CachyOS (Linux)
* **Version:** OpenDeck v2.9.1 (5290d22)
* **Installation:** Flatpak (x86_64)

aceofaces:
Plugins that update button images frequently (e.g. system monitors) send setImage with data URI payloads every second. Each call triggers save_profile(), serializing the entire profile JSON to disk — causing sustained 2-5 MB/s writes even at idle.

Data URI images are ephemeral: plugins re-push them via willAppear on startup and profile switch. Persisting them to disk is unnecessary.

This skips save_profile() when the image is a data: URI while keeping persistence for file-path images (static plugin configuration).

Fixes #249

**Preflight checklist**

---

### DIFF
--- a/src-tauri/src/events/frontend/profiles.rs
+++ b/src-tauri/src/events/frontend/profiles.rs
@@ -1,7 +1,7 @@
 use super::Error;
 
 use crate::shared::DEVICES;
-use crate::store::profiles::{PROFILE_STORES, acquire_locks_mut, get_device_profiles};
+use crate::store::profiles::{PROFILE_SAVE_DEBOUNCE, PROFILE_STORES, acquire_locks_mut, get_device_profiles, save_profile};
 
 use tauri::{AppHandle, Emitter, Manager, command};
 
@@ -29,6 +29,12 @@ pub async fn set_selected_profile(device: String, id: String) -> Result<(), Erro
 	let mut locks = acquire_locks_mut().await;
 	if !DEVICES.contains_key(&device) {
 		return Err(Error::new(format!("device {device} not found")));
+	}
+
+	// If a profile save is pending for this device, save it immediately to prevent losing profile data
+	if let Some((_, handle)) = PROFILE_SAVE_DEBOUNCE.remove(&device) {
+		handle.abort();
+		let _ = save_profile(&device, &mut locks).await;
 	}
 
 	let selected_profile = locks.device_stores.get_selected_profile(&device)?;
--- a/src-tauri/src/events/frontend/profiles.rs
+++ b/src-tauri/src/events/frontend/profiles.rs
@@ -34,7 +34,9 @@ pub async fn set_selected_profile(device: String, id: String) -> Result<(), Erro
 	// If a profile save is pending for this device, save it immediately to prevent losing profile data
 	if let Some((_, handle)) = PROFILE_SAVE_DEBOUNCE.remove(&device) {
 		handle.abort();
-		let _ = save_profile(&device, &mut locks).await;
+		if let Err(error) = save_profile(&device, &mut locks).await {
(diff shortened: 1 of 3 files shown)


nekename:
Thanks for the PR and I really appreciate the effort, but this isn't right. Plugins often do use data: URLs for less real-time changes, often when changing settings and rendering the image dynamically based on settings, and in those cases the image does need to be saved.

In order to properly address #249 (just mentioning the issue here), we'd need to use some sort of debounce, which cancels the write to the profile if another update is sent within, say, 2 seconds. This could still cause data loss if the app is quit between them but I think it's a worthwhile tradeoff.

aceofaces:
Done, Added a per-device debounce that delays the save by 2 seconds, resetting on each new call. Data may be lost if the app quits during the window, but plugins re-push images on startup via willAppear.

nekename:
Even after all my fixes it was still broken :(

If you had 1 constantly updating action it would never save the other actions, because it would always delay the save due to updates from that constantly updating action

fixed in 922d77803853e9d08083d4749913cd2c4de1e636

PR #306, opened 2026-04-09
feat: Stream Deck Plus encoder LCD, touch strip, dial input, and layouts

KatsuJinCode:
Stream Deck Plus encoder support. Rebased onto current main so this diff contains only these changes. Per-slot disk persistence split out to #312.

## Backend (Rust)

- Full 200x100 LCD rendering per encoder slot (was a 72x72 thumbnail).
- Smart resize: skip when image is already 200x100, Lanczos3 otherwise, preserving pixel-perfect plugin output.
- `setFeedback` WebSocket message support for plugin-driven LCD updates; `setFeedbackLayout` message type registered.
- Touch screen press and long-press mapped to SDK `touchTap` events. `touchTap` outbound event carries coordinates, settings, and hold state.
- `trigger_virtual_rotate`, `trigger_virtual_encoder_down`/`up`, `trigger_virtual_touch` Tauri commands for UI interaction with encoders.
- `encoder_rotated` frontend event for real-time dial tracking from physical device.

## Frontend (Svelte)

- Edge-to-edge encoder strip matching physical 800x100 LCD layout.
- Correct corner rounding (rounded outer edges, square inner edges).
- Interactive `EncoderDial` component with 30-position notch indicator.
- Dial / Touch Press and Hold controls with visual feedback.
- Keyboard shortcut binding for all encoder actions.
- Physical device rotation reflected in UI dial indicator.
- Dark theme CSS for Property Inspector select elements.


KatsuJinCode:
## Additional feature: per-slot disk persistence control

Building on your debounce approach in #302, we added user-level control over image persistence:

### The problem we both found independently
Plugins that update frequently (system monitors, visualizers) trigger continuous disk writes via \`save_profile()\`. You solved this with debouncing for data URI images. We initially solved it by removing the writes entirely. Both are compromises -- debouncing still writes periodically, full removal means nothing survives restart.

### The solution: let the user decide per-slot
- **Right-click context menu**: "Skip disk writes" checkbox on any assigned slot. Toggling it sets an explicit override. Toggling again resets to follow the global default.
- **Global default in Settings**: "Skip image persistence by default" -- when enabled, all slots skip persistence unless individually overridden.
- **Flag on ActionInstance**: \`skip_persistence: Option<bool>\` -- \`None\` = follow global, \`Some(true/false)\` = explicit override. Travels with the instance, cleared on new instance creation, survives restarts.

When persistence is skipped for a slot, \`setImage\` and \`setTitle\` still render on the device in real time -- they just don't write to disk. Plugins re-push images via \`willAppear\` on restart.

This gives users running performance monitors zero disk I/O without affecting users who want normal persistence. No heuristics, no guessing -- the user knows which buttons are live monitors and which should persist.

### Files changed in this addition
- \`src-tauri/src/shared.rs\` -- \`skip_persistence: Option<bool>\` on ActionInstance
- \`src-tauri/src/store/mod.rs\` -- \`skip_persistence_default\` in Settings
- \`src-tauri/src/store/simplified_profile.rs\` -- preserve flag through disk conversion
- \`src-tauri/src/events/inbound/states.rs\` -- check flag in set_image/set_title
- \`src-tauri/src/events/frontend/instances.rs\` -- toggle_skip_persistence command
- \`src/components/Key.svelte\` -- context menu checkbox
- \`src/components/SettingsView.svelte\` -- global default toggle

Terrorwolf01:
Please make separate PRs for different Features and not everything into one.

KatsuJinCode:
Did you want me to resubmit this one or just for the future?

nekename:
Please fix the base of this PR as the diff is unreviewable in its current state

nekename:
> Did you want me to resubmit this one or just for the future?

Sorry, your comment hadn't loaded when I sent my previous comment. Use an interactive rebase to drop the commits irrelevant to Stream Deck Plus support. Then, rebase onto current main so that the diff in this PR only contains your changes.

KatsuJinCode:
Rebased onto current \`main\` and split per feedback. The diff here now contains only the Stream Deck Plus encoder LCD + touch strip + dial changes.

The per-slot disk persistence work is now a separate PR: #312.

Let me know if you'd like Stream Deck Plus support split further (e.g. LCD rendering / touch strip / virtual dial controls as separate PRs) -- I kept them together because they share the same files and changes are tightly coupled, but I'm happy to split if it helps review.

Terrorwolf01:
Hi,
Please also put the Virtual Encoder in its own PR. Please also add, that as long as there is no support for Layouts, that the image will be a square and not stretched to the limit. Also, since there is no Top, please remove the Top Indicator on the Encoders shown in OpenDeck.
May I ask whats your reasoning behind changing the Space where actions are to be dropped form the circle to the Strip?
The official Software also uses the Encoder circle for this.

KatsuJinCode:
Update on @Terrorwolf01's feedback:

**Virtual encoder split**: done, opened as #313 (depends on this one).

**Built-in layout support**: implemented in this PR. The setFeedback / setFeedbackLayout handlers now track per-instance layout + feedback state and dispatch to a frontend renderer. All 6 built-in layouts (\$X1, \$A0, \$A1, \$B1, \$B2, \$C1) plus arbitrary custom layout JSON files from plugin folders. Renderer covers the four item types (pixmap, text, bar, gbar), all 5 bar subtypes, gradient color specs, z-order, opacity, text alignment + clip/ellipsis overflow.

Tested on actual Stream Deck Plus hardware:
- All six built-in layouts ($X1, $A0, $A1, $B1, $B2, $C1) render correctly when cycling them via the official `elgatosf/streamdeck-plugin-samples` `layouts` plugin's "Built-in Layout" action.
- Custom layout JSON loading + text-heavy rendering + gbar indicator validated end-to-end with the Adaptive Dial Scrolling demo plugin.
- Dial input, touch, setFeedback updates from the plugin all reach the device LCD correctly.

**Image rendering**: native 200x100 per encoder, no stretching. Pixmap items inside layouts preserve aspect ratio; only the `full-canvas` key in $A0 (which the SDK docs explicitly define as 200x100) is stretched to fit.

**Drop target / circle vs strip**: the physical Stream Deck Plus has a single 800x100 LCD strip across all four encoders. The strip is what the device displays; the dial is the input. They aren't the same thing, and rendering the screen as a circle doesn't reflect what's on the hardware. OpenDeck's UI matches the device.

Also fixed an unrelated upstream issue while in this code: PNG was missing from the `image` crate's enabled features in `Cargo.toml`. The encoder LCD path (and any plugin sending PNG via setImage) was hitting "image format Png is not supported" warnings in the log. Added `png` to the features list.

Terrorwolf01:
Hi,
So. Still multiple issues.
Non Layout Actions are still getting stretched.
For the Layouts: Its flickering and also stops working after a few seconds.
Also, You didn't answer my Question about the Dragging Actions to the encoder instead of the screen. It seems that its just an answer of an AI which didn't get the right prompt. You are assigning an action to an encoder. Not the screen. So this should stay in the current OpenDeck Behavior. The screen still needs to be rendered in OpenDeck tho. Excatly like in the official Elgato Software.
Also the Layouts and the Touch Part maybe should be separated into diferent PRs to be able to better review it.
Also the fix you mentioned in your last comment, again, should be a separate PR aswell.

KatsuJinCode:
On Wed, Apr 15, 2026, 10:53 AM Terrorwolf01 ***@***.***>
wrote:

> *Terrorwolf01* left a comment (nekename/OpenDeck#306)
> <https://github.com/nekename/OpenDeck/pull/306#issuecomment-4254269767>
>
> Hi,
> So. Still multiple issues.
> Non Layout Actions are still getting stretched.
>
Can you tell me more, it's working perfectly for me without stretching.

> For the Layouts: Its flickering and also stops working after a few seconds.
>
See above, tested and working.

> Also, You didn't answer my Question about the Dragging Actions to the
> encoder instead of the screen. It seems that its just an answer of an AI
> which didn't get the right prompt. You are assigning an action to an
> encoder. Not the screen. So this should stay in the current OpenDeck
> Behavior.
>
I'm confused by what you're saying here yes the first party el gato stream
deck application is ergonomically deficient in many ways my aim was to
improve upon it.

Thee screen still needs to be rendered in OpenDeck tho. Excatly like in the
official Elgato Software.

Tested and working.

> Also the Layouts and the Touch Part maybe should be separated into
> diferent PRs to be able to better review it.
> Also the fix you mentioned in your last comment, again, should be a
> separate PR aswell.
>
Do you want me to keep submitting PRs to you...? Should I maintain a fork
and you can pick and choose what if anything you want?

Terrorwolf01:
We are happy for, PRs, just its easier to review the code, if things are Properly separated.

> Can you tell me more, it's working perfectly for me without stretching.
image


> Can you tell me more

Happy to jump in a Discord Call or something to show and help Debug.

>I'm confused by what you're saying here yes the first party elgato stream deck application is ergonomically deficient in many ways my aim was to improve upon it.

Same as above, happy to hop in a call to discuss a bit. As said before, in the Technical side, you are not assigning actions to the Screen, but to the Encoder.
So in my Personal opinion, this should be reflected in the UI.
A Side effect of this would be, that we have the screen, to be able to configure Swipe Actions which are blocked without major rewrites of the Frontend if you use it to configure the Encoders.

KatsuJinCode:
Are those encoder actions? They look like standard key plugin actions,
could have a second path to keep non-native encoder actions unstretched
when dropped on encoders.

The encoder and screen are a single unit it's not possible to assign
something separately to each so while you're correct that the encoder is
being assigned, since the screen is the main interface for most people I
mean for many people just the concepts of an app exists as purely the front
end feedback to me it's very strange to drag an action to the dial itself
let alone to have it appear on the dial maybe it's just a me thing but it
seems way more intuitive, though supporting dropping on either element
should be trivial.

Swiping logic should be simple enough, I'll have to test it on the device
to see how sensitive the digitizer is and exact timings for ergonomics but
I can get around to it shortly.

Terrorwolf01:
> Are those encoder actions?

Yes. In the Plugin Manifest, its defined as Encoder Actions. In the ActionList, there is no visual distinction. I am working behind the scene on something regarding this tho.

As for the Swiping Actions: Since we don't have Pages like in the official App, we need a configuration to say to which Profile one wants to swipe. Thats what the Visual Bar should be used in my Opinion instead of the Encoder Actions.

KatsuJinCode:
This batch should be more manageable, I had it split up and better
commented. Hopefully it's not so much work for you, apologies for that.

KatsuJinCode:
Pushed fixes for the issues you reported:

**Stretched non-layout actions**: Actions without an explicit `Encoder.layout` in their manifest now default to the `$X1` (Icon) layout instead of being stretched to 200x100. This applies both to the frontend layout resolution and the backend instance creation.

**Flickering / stops working**: Three fixes:
1. Encoder LCD preview now renders to an offscreen canvas and blits atomically, preventing flash during async image decode.
2. Null image pushes (`update_image(null)`) are blocked for encoder controller slots. These were being sent during profile switches, racing with real `setFeedback` content from plugins.
3. Added a backend fast path for `setFeedback` with a `full-canvas` data URI -- pushes directly to the device LCD from the backend, bypassing frontend render latency.

**Drag target (encoder dial vs strip)**: Fair point about assigning to the encoder, not the screen. I'll look at supporting both as drop targets. The strip rendering needs to stay since that's what the physical device displays.

**Separate PRs**: The PNG fix has been removed from this PR (will submit separately). Virtual encoder controls were already split to #313. Layouts and touch handling share the same files and state (`feedback_layout`, `feedback`, `setFeedback`, `setFeedbackLayout`) so splitting them would create interdependent PRs that can't be reviewed or tested independently.

KatsuJinCode:

nekename:

PR #312, opened 2026-04-15
feat: per-slot disk persistence control with global default

KatsuJinCode:
## Motivation

Plugins that update frequently (system monitors, visualisers) trigger continuous disk writes via \`save_profile\`. The existing debounce on data-URI images in \`set_image\` helps, but still writes periodically and does nothing for non-data-URI image paths or \`setTitle\`.

This complements #302 rather than replacing it: debounce still protects the common case; this adds an explicit opt-out for plugins the user knows are live-updating.

## What this adds

- **Per-slot toggle.** Right-click context menu on any assigned slot: "Skip disk writes" checkbox. Cycles \`None -> Some(!default) -> None\` so toggling twice resets to "follow global default" instead of being permanently overridden.
- **Global default.** "Skip image persistence by default" toggle in Settings. Applies to all slots unless individually overridden.
- **\`skip_persistence: Option<bool>\`** on \`ActionInstance\`. \`None\` = follow global, \`Some(true/false)\` = explicit override. Preserved through \`DiskActionInstance\` conversion so it survives restart.
- \`set_image\` and \`set_title\` check the flag and skip \`save_profile\` when set. Device still renders in real time via \`update_image\`; only disk I/O is skipped. Plugins re-push images via \`willAppear\` on restart, so skipped slots recover their visual state without persistence.

## Files

- \`src-tauri/src/shared.rs\` -- field on \`ActionInstance\`
- \`src-tauri/src/store/mod.rs\` -- global default in \`Settings\`
- \`src-tauri/src/store/simplified_profile.rs\` -- preserve flag through disk conversion
- \`src-tauri/src/events/inbound/states.rs\` -- check flag in \`set_image\` / \`set_title\`
- \`src-tauri/src/events/frontend/instances.rs\` -- \`toggle_skip_persistence\` command
- \`src-tauri/src/main.rs\` -- register command
- \`src/components/Key.svelte\` -- context menu checkbox
- \`src/components/SettingsView.svelte\` -- global default toggle

### DIFF
--- a/src-tauri/src/events/frontend/instances.rs
+++ b/src-tauri/src/events/frontend/instances.rs
@@ -29,6 +29,7 @@ pub async fn create_instance(app: AppHandle, action: Action, context: Context) -
 			current_state: 0,
 			settings: serde_json::Value::Object(serde_json::Map::new()),
 			children: None,
+			skip_persistence: None,
 		};
 		children.push(instance.clone());
 
@@ -59,6 +60,7 @@ pub async fn create_instance(app: AppHandle, action: Action, context: Context) -
 			} else {
 				None
 			},
+			skip_persistence: None,
 		};
 
 		*slot = Some(instance.clone());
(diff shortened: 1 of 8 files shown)


nekename:
This really doesn't seem like the correct approach, considering I know of no such option in Elgato's software. I am going to hold off on solving this issue thoroughly until we have both a good understanding of an approach that keeps this hidden from the user, and a larger-scale refactor of the profile system to reduce temperamental writes.

PR #313, opened 2026-04-15
feat: virtual encoder controls (click/drag dial widget and touch simulation)

KatsuJinCode:
Per @Terrorwolf01's request to split the virtual encoder out: this PR adds a UI-only layer on top of #306 that lets users interact with the encoder from OpenDeck itself, without needing the physical device.

Depends on #306 (the touch_tap handler used by `trigger_virtual_touch` lives there).

## Frontend (Svelte)

- `EncoderDial` component per encoder: 30-detent visual indicator, click-drag to rotate, click to press / release, long-press for touch-and-hold, double-click for touchTap.
- Physical device rotation is reflected in the UI dial via the `encoder_rotated` event, so the indicator tracks real hardware too.

## Backend (Rust)

- `trigger_virtual_rotate`, `trigger_virtual_encoder_down` / `_up`, `trigger_virtual_touch` Tauri commands. They feed the same inbound handlers as real device events, so plugins see no difference between virtual and physical input.
- `encoder_rotated` frontend event emitted from `encoder_change` so the dial indicator can track physical rotation in real time.
- `TicksPayload` made `Serialize + Clone` so it can be emitted to the webview.

## Why split this out

@Terrorwolf01 noted the official Stream Deck software uses a circular dial UI for encoder slots; this PR gives OpenDeck the equivalent. The visible LCD strip rendering (PR #306) is independent of this -- you can ship that without this PR if you'd prefer to keep OpenDeck's encoder UX more minimal.


nekename:
I'm sorry, but there is no way I will be able to review or merge a PR that changes over 1,300 lines of code.

nekename:

nekename:

PR #314, opened 2026-04-16
fix: kill plugin child processes on exit for all platforms

KatsuJinCode:
## Summary

- `deactivate_plugins()` was gated behind `#[cfg(windows)]`, so on Linux and macOS every plugin child process (Node.js, Wine, native) became an orphan when OpenDeck exited
- After multiple restarts these accumulated into dozens of zombie processes consuming gigabytes of RAM
- Removed the platform gate so all platforms clean up plugin processes on exit via the existing `deactivate_plugin()` mechanism (`child.kill()` + `child.wait()`)

## Changes

**`src-tauri/src/main.rs`** — Removed `#[cfg(windows)]` from the `deactivate_plugins()` call in the `RunEvent::Exit` handler.

**`src-tauri/src/plugins/mod.rs`** — Removed `#[cfg(windows)]` from the `deactivate_plugins()` function definition.

## Test plan

- [ ] Start OpenDeck on Linux, verify plugins start (check `pgrep -af sdPlugin`)
- [ ] Close OpenDeck, verify all plugin processes are gone (same command shows nothing)
- [ ] Restart OpenDeck multiple times, confirm no orphaned processes accumulate

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -364,7 +364,9 @@ If you have already donated, thank you so much for your support!"#,
 
 	app.run(|app, event| {
 		if let tauri::RunEvent::Exit = event {
-			#[cfg(windows)]
+			// Kill plugin child processes on exit. Without this, every Node.js /
+			// Wine / native plugin process becomes an orphan when OpenDeck shuts
+			// down on Linux and macOS, accumulating across restarts.
 			futures::executor::block_on(plugins::deactivate_plugins());
 			tokio::spawn(elgato::reset_devices());
 			use tauri_plugin_aptabase::EventTracker;
--- a/src-tauri/src/plugins/mod.rs
+++ b/src-tauri/src/plugins/mod.rs
@@ -362,7 +362,8 @@ pub async fn deactivate_plugin(app: &AppHandle, uuid: &str) -> Result<(), anyhow
 	}
 }
 
-#[cfg(windows)]
+/// Terminate all running plugin child processes. Called on application exit
+/// to prevent orphaned Node.js / Wine / native processes from accumulating.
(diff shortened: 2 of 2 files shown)


nekename:
Thanks for the contribution! My concern with this approach, and in fact why this was gated to Windows in the first place, is that some plugins rely on being able to terminate themselves gracefully and/or intercept SIGTERM to be able to handle cleanup logic, for example resetting a hardware device.

I have opened a PR to elgatosf/streamdeck to solve this problem on the side of Node.js plugins, however. Could you let me know which specific plugins you're encountering this with, and if any of them aren't Node.js-based, then we can fix the issue in any other plugin SDKs?

KatsuJinCode:
The orphan process issue specifically affects **Node.js plugins** on Linux. When OpenDeck exits, Node.js child processes don't terminate and accumulate across restarts (each consuming 150-280MB). Native Rust plugins (e.g. the starterpack) exit correctly with the parent process.

All of our custom plugins are Node.js-based: system-monitor, claude-monitor, custom-radial-menu, touch-button-test, encoder-multi-key, ultimate-playback-control, ultimate-playback-youtube, ultimate-custom-mcp-utility, adaptive-scrolling-demonstration.

Updated the PR to address the graceful shutdown concern: plugins now receive SIGTERM first with a 3-second grace period for cleanup, falling back to SIGKILL only if the process doesn't exit. This preserves the ability for plugins to intercept SIGTERM and run cleanup logic (resetting hardware, flushing state, etc.).

nekename:
The three second grace period isn't a good solution, because you haven't even made it multithreaded, so 10 blocking plugins amounts to a 30 second delay. I'm going to close this and await the fix to be merged upstream

PR #315, opened 2026-04-16
feat: encoder layout rendering improvements

KatsuJinCode:
## Summary

Follow-up to #306 addressing feedback from @Terrorwolf01 about stretched images and flickering on encoder LCDs. These changes improve how encoder layout rendering works without changing the core encoder/touch/dial support from #306.

- **Default $X1 layout fallback**: when an encoder action's manifest omits `Encoder.layout`, default to the built-in `$X1` ("Icon") layout instead of stretching the action image to the full 200×100 LCD. This fixes the stretched non-layout encoder actions @Terrorwolf01 reported.

- **Manifest refresh on profile load**: when OpenDeck loads a saved profile, it now refreshes each slot's cached `action` field from the current plugin manifest. This heals slots that were saved before a plugin added `Encoder.layout` to its manifest -- the layout picks up automatically on next launch without needing to remove and re-add the action.

- **setFeedback fast path for full-canvas plugins**: plugins that pre-render their own 200×100 PNG and push it via `setFeedback({ "full-canvas": dataUri })` now have the image sent directly to the device from the backend, bypassing the webview compositor round-trip. This gives the same latency as `setImage` while using the correct SDK path. Plugins that push partial updates (title, bar, indicator) still use the frontend compositor as before.

- **Flash prevention on profile switch**: blocks redundant null-image pushes to encoder LCD slots from the frontend. Profile switches already use `clear_screen` to blank the device; individual null pushes from slot teardown were racing with the fast path, causing visible flashing.

- **Offscreen preview composition**: encoder LCD previews in the app window are rendered to an offscreen canvas and blitted in one operation, eliminating the visible black-frame flicker during async pixmap decode.

## Changes

| File | What changed |
|------|-------------|
| `instances.rs` | `default_encoder_layout` closure defaults to `$X1` for encoder slots; `update_image` drops null pushes for encoder controller |
| `profiles.rs` | `refresh_instance` updates slot's `action` from manifest and heals null `feedback_layout` |
| `states.rs` | Fast path in `set_feedback`: detects `full-canvas` data URI, pushes to device directly with profile check |
| `will_appear.rs` | Warming-up flag: marks encoder contexts after `willAppear`, cleared on first `setFeedback` to prevent stale-frame flash |
| `Key.svelte` | `$X1` fallback in layout resolution; offscreen compose for fast-path preview; skip `renderImage` fallback and null push for encoder slots |

## Test plan

- [ ] Place an encoder action without `Encoder.layout` in manifest on an encoder slot -- should show centered icon via $X1, not stretched
- [ ] Plugin with `Encoder.layout: "$A0"` that pushes `setFeedback({ "full-canvas": png })` -- should render at full speed on LCD with no webview delay
- [ ] Switch profiles via the UI -- encoder LCDs should clear and reload without flashing
- [ ] Rapid profile switching back and forth -- no sustained flashing or seizure-mode flickering


nekename:

KatsuJinCode:
Thanks - marked as draft.

KatsuJinCode:

nekename:

PR #316, opened 2026-04-16
feat: drag-drop UX improvements

KatsuJinCode:
## Summary

Improves the action drag-and-drop experience with visual feedback and usability fixes.

- **Icon-only drag ghost**: dragging an action from the sidebar shows just the icon instead of the full card with title text, so it doesn't obscure the slot you're dropping onto
- **Drop on occupied slots**: previously silently rejected; now overwrites the existing action (removes old, places new)
- **Visual drop-target highlighting**: slots light up during drag to show compatibility and hover state
  - 🟢 Green border: empty compatible slot
  - 🟠 Orange border: occupied compatible slot (dropping will overwrite)
  - 🔵 Blue border + brightness boost: currently hovered slot
  - ⬛ Dimmed: incompatible slot (action doesn't support that controller type, e.g. dragging an Encoder-only action dims all Keypad slots)
- **Combined encoder drop target**: the encoder strip and its associated area highlight as one unit per encoder position
- **Selection clears on drag start**: reduces visual noise during drag

## Changes

| File | What changed |
|------|-------------|
| `ActionList.svelte` | `setDragImage` for icon-only ghost; publish `dragAction` store on drag start/end; clear selection |
| `DeviceView.svelte` | `handleDragOver` tracks hovered slot; `handleDrop` allows overwrite via `remove_instance`; computed `dragHighlight` per slot; combined encoder container with highlight classes |
| `Key.svelte` | `dragHighlight` prop; colored border classes on encoder strip and keypad canvases; `on:dragleave` forwarding |
| `dragState.ts` | New shared stores: `dragAction`, `hoveredSlot`, `openDialPosition` |

## Test plan

- [ ] Drag action from sidebar -- ghost should be icon only, not full card
- [ ] Drag over empty compatible slot -- green border
- [ ] Drag over occupied compatible slot -- orange border
- [ ] Hover a slot -- blue border, brighter
- [ ] Drag encoder-only action -- keypad slots dim, encoder slots highlight
- [ ] Drop on occupied slot -- old action replaced
- [ ] Drop on incompatible slot -- nothing happens (dimmed, no drop effect)


nekename:
This PR seems to contain the entire diff for the Stream Deck Plus PR as well. Is the intention that this is to be merged after that one?

Anyway, the current behaviour is intentional. The full card being dragged was in fact implemented after-the-fact to make the dragging experience more cohesive regardless of whether you dragged the icon or the text. Silent rejection when dragging to occupied slots is preferred over replacement. I'm not sure if I like the colour coding either, it's too distracting.

Please simply add the changes for unifying encoder drop targets to #306.

PR #317, opened 2026-04-16
feat: touch strip swipe to switch profiles

KatsuJinCode:
## Summary

Wires the previously-unused `TouchScreenSwipe` HID event to cycle between profiles on Stream Deck+ devices.

- **Any horizontal swipe** on the touch strip switches profiles: swipe right = next profile alphabetically, swipe left = previous, wrapping at both ends
- Uses the **full profile switch path** (`willDisappear` / `clear_screen` / `willAppear`) so plugins are properly notified and the device LCD is cleared before loading the new profile
- The `TouchScreenSwipe` event was already parsed by the `elgato_streamdeck` HID crate but mapped to `Ok(())` (no-op). This PR connects it to a handler

On the official Elgato software, swiping on the touch strip navigates between pages. OpenDeck uses profiles instead of pages, so the swipe cycles through all profiles on the device.

## Changes

| File | What changed |
|------|-------------|
| `elgato.rs` | Wire `TouchScreenSwipe(from, to)` to `touch_swipe` handler (1 line) |
| `devices.rs` | New `touch_swipe` function: computes swipe direction from dx, lists profiles alphabetically, calls `set_selected_profile` + emits `switch_profile` to frontend |
| `misc.rs` | Made `SwitchProfileEvent` fields pub + added `new()` constructor so it can be created from `devices.rs` |

## Test plan

- [ ] Create 2+ profiles on a Stream Deck+ device
- [ ] Swipe right on touch strip -- profile switches to next
- [ ] Swipe left -- profile switches to previous
- [ ] Swipe on last profile wraps to first (and vice versa)
- [ ] Encoder LCD clears cleanly on switch, plugins receive willDisappear/willAppear
- [ ] Quick repeated swipes cycle through all profiles without freezing


nekename:

KatsuJinCode:
Thanks - marked as draft.

KatsuJinCode:

nekename:

PR #318, opened 2026-04-19
fix: protect plugin children from orphaning via PR_SET_PDEATHSIG on Linux

KatsuJinCode:
## Problem

When OpenDeck exits via SIGKILL, panic, or any path that doesn't run the graceful `deactivate_plugins` cleanup added in #314, plugin subprocesses aren't signaled. On Linux they reparent to PID 1 / user-systemd and keep running indefinitely — the classic orphan leak. Each orphaned plugin holds its WebSocket client (reconnecting in a loop), its polling loops, and its full heap. Across multiple OpenDeck restarts during development or instability, orphans accumulate with no bound.

Reproducing on current `main`:
```
pkill -9 opendeck          # simulate crash / force-kill
pgrep -af -- '-pluginUUID' # every previously-running plugin survives
```

## Fix

Set `PR_SET_PDEATHSIG=SIGTERM` on every plugin child via `CommandExt::pre_exec`. The kernel delivers SIGTERM to the child when its parent (this OpenDeck) dies, regardless of whether `deactivate_plugins` ran. Applied to all three non-Windows spawn sites (Node, Wine, Native).

After the fix:
```
pkill -9 opendeck
sleep 2
pgrep -af -- '-pluginUUID' # empty
```

## Relationship to #314

Complementary. #314 handles the graceful-exit case — SIGTERM delivered to each plugin via `deactivate_plugins`, with a 3-second grace window before SIGKILL. This PR handles the non-graceful-exit case — crash, panic, external SIGKILL. Together they cover both ends.

## Platform scope

Linux-only. `PR_SET_PDEATHSIG` is a Linux-specific prctl; no macOS or Windows equivalent.

- macOS: equivalent is `kqueue` with `NOTE_EXIT` on the parent PID, requires a monitor thread in each child.
- Windows: equivalent is `JobObject` with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, set on the parent before spawning children.

Both are follow-up work.

## Safety

`pre_exec` runs in the forked child between `fork()` and `execve()`. Only async-signal-safe calls are permitted inside the closure. `prctl` is async-signal-safe per `signal-safety(7)`.

## Dependencies

Adds `libc = "0.2"` under a Linux-only target dep.

## Testing

Verified manually:
1. Apply patch, rebuild, install.
2. Start OpenDeck, confirm plugins spawned.
3. \`kill -9 \$(pgrep -x opendeck)\` — all plugin processes exit within 2 seconds.
4. Without the patch, same procedure leaves every plugin running as orphan.

No behavior change on macOS or Windows (helper is a no-op stub on non-Linux).

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -3375,6 +3375,7 @@ dependencies = [
  "hidapi",
  "image",
  "json-patch 4.1.0",
+ "libc",
  "log",
  "log-panics",
  "open",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -52,6 +52,9 @@ path-slash = "0.2"
 fix-path-env = { git = "https://github.com/tauri-apps/fix-path-env-rs", rev = "c4c45d503ea115a839aae718d02f79e7c7f0f673" }
 font-loader = "0.11"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libc = "0.2"
+
 [target.'cfg(windows)'.dependencies]
 windows-sys = "0.61"
 
--- a/src-tauri/src/plugins/mod.rs
+++ b/src-tauri/src/plugins/mod.rs
@@ -45,6 +45,21 @@ pub static PORT_BASE: LazyLock<u16> = LazyLock::new(|| {
 	base
 });
 
+/// Attach a kernel-enforced "die when parent dies" signal to a plugin child process.
+#[cfg(target_os = "linux")]
+fn attach_parent_death_signal(command: &mut Command) {
+	use std::os::unix::process::CommandExt;
+	// SAFETY: `libc::prctl` is async-signal-safe.
+	unsafe {
+		command.pre_exec(move || {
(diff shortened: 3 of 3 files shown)


nekename:
This looks like a cleaner solution than #314 that I'd be happy to merge, if it works. Waiting on Copilot review and my own testing. There's a couple changes I'll need to make to the implementation details but the idea seems robust

KatsuJinCode:
Pushed 509603e addressing both review points.

Vararg cast: applied as suggested.

For the race window, deviated from the suggested `getppid() == 1` pattern. The textbook check only catches orphans that reparent directly to PID 1, but on any Linux system where an ancestor has called `PR_SET_CHILD_SUBREAPER` — notably the per-user `systemd --user` instance on most modern desktops, including the system this PR was developed on — orphans reparent to the sub-reaper instead, and the textbook check silently misses the transition.

The fix here captures the parent's pid before spawn and compares the child's `getppid()` against that value. Robust regardless of where in the reaper hierarchy the orphan ends up.

nekename:
Thanks @KatsuJinCode and @claude

PR #319, opened 2026-04-19
feat: activate plugins per profile, deactivate on settle

KatsuJinCode:
## Problem

OpenDeck currently spawns every installed plugin at startup regardless of whether any profile references it. This creates three compounding issues:

1. **Idle-plugin memory pressure.** Plugins the user doesn't have any action for still run continuously. Polling loops, WebSocket clients, per-plugin heaps — all active.
2. **The problem gets much worse with #317.** Touch-strip swipe to switch profiles makes profile switches cheap and frequent. Without lazy activation, every profile's plugins stay resident simultaneously. Users who swipe through 5 profiles have every plugin from all 5 running, even though only one profile is on screen.
3. **Divergence from Elgato's real Stream Deck.** Elgato's host launches plugins on demand based on the actions the user has configured, not eagerly.

## Change

- **Startup:** register every installed plugin's metadata (`CATEGORIES`, `DEVICE_NAMESPACES`) so the profile editor shows available actions for all installed plugins, but only spawn subprocesses for plugins referenced by at least one profile on disk.
- **Profile switch / device connect:** lazily spawn plugins the new profile references that aren't already running (`ensure_plugin_spawned`).
- **Deactivation on settle:** 30-second debounce after any profile/device event. Once settled, deactivate any plugin not referenced by the selected profile on any connected device. Rapid profile flipping cancels and restarts the timer — no plugin thrashing during fast swipes.

Matches the lazy-activation behavior of Elgato's real Stream Deck.

## Why not keep plugins warm

Rejected as the wrong default. If a plugin isn't referenced by the current profile on any connected device, the user isn't using it. Keeping it resident for some arbitrary N-last-used window trades clarity for an optimization that doesn't matter — plugin startup is fast enough that users won't notice. Every plugin is already restart-safe by construction (OpenDeck kills and respawns every plugin on every OpenDeck restart; a plugin that can't survive that is dead on day 1).

## Atomicity fix bundled

`deactivate_plugin` previously used `?` to short-circuit out of namespace cleanup. If `deregister_device` returned `Err`, the plugin's subprocess would stay running while its namespace was already removed from `DEVICE_NAMESPACES`. Subsequent sweeps couldn't find it cleanly. This PR wraps the namespace cleanup in an inner async block, logs any error via `warn!`, and always proceeds to the subprocess-kill path.

## Concurrent-spawn race guard

`initialise_plugins` fires parallel `tokio::spawn(initialise_plugin)` per plugin directory. If a device registers during that window, `ensure_plugin_spawned` for a UUID that's already in-flight would start a **second** subprocess; the second's `INSTANCES.insert` overwrites the first, and the first's `Child` handle is dropped — which on Unix does NOT kill the process. Result: a self-inflicted orphan.

This PR adds `SPAWNING_UUIDS: std::sync::Mutex<HashSet<String>>` with `try_claim_spawn` (atomic check-and-insert via `HashSet::insert` return value) and a `SpawnClaimGuard` RAII wrapper that releases the claim on any exit from the spawn path (return, `?`, panic unwind).

## Platform behavior

Behavior identical on all platforms. No OS-specific code.

## Testing

1. **Startup filter:** with 11 installed plugins, 3 referenced by the active profile. On current `main`: 11 subprocesses. With this PR: 3 subprocesses, 8 log lines reading `Registering X metadata only (not referenced by any profile)`. Profile editor UI still shows all 11 plugins' actions.
2. **Profile switch lazy spawn:** switch to a profile that references an unstarted plugin. That plugin's subprocess appears within the `will_appear` event firing.
3. **Deactivation on settle:** switch to a profile that doesn't reference a currently-running plugin. 30 seconds later, log shows `Deactivation sweep: removing N unused plugin(s)` and `pgrep -af -- '-pluginUUID'` drops.
4. **Rapid profile flipping:** switch A → B → A within 5 seconds. No deactivation fires (timer cancelled and restarted each time). Plugin count stable.
5. **Concurrent spawn race:** Hard to trigger naturally; covered by structural fix + RAII guard.
6. **Atomicity:** forced a `deregister_device` error via a contrived device namespace; the plugin subprocess was still killed, namespace error logged.

## Reviewer context

This is a behavioral change to plugin lifecycle. Happy to split further if you'd prefer the concurrent-spawn race fix and atomicity fix as their own PR — they're defensive hardening that landed in the same commit because the lazy-spawn path exposed them. Let me know.

### DIFF
--- a/src-tauri/src/events/frontend/plugins.rs
+++ b/src-tauri/src/events/frontend/plugins.rs
@@ -105,14 +105,14 @@ pub async fn install_plugin(app: AppHandle, url: Option<String>, file: Option<St
 	if let Err(error) = crate::zip_extract::extract(std::io::Cursor::new(bytes), &config_dir.join("plugins")) {
 		log::error!("Failed to unzip file: {}", error);
 		let _ = fs::rename(&temp, &actual).await;
-		let _ = crate::plugins::initialise_plugin(&actual).await;
+		let _ = crate::plugins::initialise_plugin(&actual, true).await;
 		return Err(anyhow::Error::from(error).into());
 	}
-	if let Err(error) = crate::plugins::initialise_plugin(&actual).await {
+	if let Err(error) = crate::plugins::initialise_plugin(&actual, true).await {
 		log::warn!("Failed to initialise plugin at {}: {}", actual.display(), error);
 		let _ = fs::remove_dir_all(&actual).await;
 		let _ = fs::rename(&temp, &actual).await;
-		let _ = crate::plugins::initialise_plugin(&actual).await;
+		let _ = crate::plugins::initialise_plugin(&actual, true).await;
 		return Err(error.into());
 	}
 	let _ = fs::remove_dir_all(config_dir.join("temp")).await;
@@ -153,7 +153,7 @@ pub async fn remove_plugin(app: AppHandle, id: String) -> Result<(), Error> {
 #[command]
 pub async fn reload_plugin(app: AppHandle, id: String) {
 	let _ = crate::plugins::deactivate_plugin(&app, &id).await;
-	let _ = crate::plugins::initialise_plugin(&config_dir().join("plugins").join(&id)).await;
+	let _ = crate::plugins::initialise_plugin(&config_dir().join("plugins").join(&id), true).await;
 
 	let locks = acquire_locks().await;
 	let all = locks.profile_stores.all_from_plugin(&id);
--- a/src-tauri/src/events/frontend/profiles.rs
+++ b/src-tauri/src/events/frontend/profiles.rs
@@ -67,6 +67,24 @@ pub async fn set_selected_profile(device: String, id: String) -> Result<(), Erro
 	// We must use the mutable version of get_profile_store in order to create the store if it does not exist.
 	let store = locks.profile_stores.get_profile_store_mut(&DEVICES.get(&device).unwrap(), &id).await?;
 	let new_profile = &store.value;
+
+	// Lazy plugin activation: collect the plugin UUIDs referenced by the new
+	// profile and ensure each has a running subprocess before firing
+	// will_appear events. Plugins registered metadata-only at startup
+	// (unreferenced) are spawned here on first use.
+	let mut needed_plugins = std::collections::HashSet::<String>::new();
+	for instance in new_profile.keys.iter().flatten().chain(&mut new_profile.sliders.iter().flatten()) {
+		needed_plugins.insert(instance.action.plugin.clone());
+		if let Some(children) = &instance.children {
+			for child in children {
+				needed_plugins.insert(child.action.plugin.clone());
+			}
+		}
+	}
+	for uuid in &needed_plugins {
+		crate::plugins::ensure_plugin_spawned(uuid).await;
+	}
+
 	for instance in new_profile.keys.iter().flatten().chain(&mut new_profile.sliders.iter().flatten()) {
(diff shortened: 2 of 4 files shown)


Terrorwolf01:
2 Questiosn:
Do you have anything to back up that the official Software does that? Last time I checked, the official Software had all Plugins running regardless if an Action was configured or not.
Did you happen to test this with a DevicePlugin where you have a non-Elgato Device added, so it doesn't kill the device Plugins?

KatsuJinCode:
> Do you have anything to back up that the official Software does that?

Yes. Direct observation — I've run Stream Deck with 50+ plugins installed; the host doesn't have 50+ plugin subprocesses resident on the machine. Anyone with a meaningfully loaded setup sees the same thing in their process list.

The SDK architecture is also only coherent under demand-based activation:

- `onWillAppear` and `onWillDisappear` are how a plugin learns which action instances exist in the current context. The events are scoped to actions currently visible in the active profile / page / folder. A plugin whose actions aren't in any visible context receives no events and has no state to maintain — running its subprocess serves no purpose.
- The SDK describes "lifecycle management" with "automatic failure recovery" — language that only applies to a system that starts and stops subprocesses. An architecture that spawned every installed plugin at startup and left them running wouldn't need either concept.

If plugins were always resident regardless of use, `willAppear` / `willDisappear` wouldn't need to be scoped to visibility — the plugin could just be told about every action instance in every profile once on connect. The event model is shaped the way it is because the subprocess lifecycle is driven the same way.

OpenDeck currently spawns every installed plugin regardless; this PR brings it to parity with the actual behavior of the host.

> Did you happen to test this with a DevicePlugin where you have a non-Elgato Device added, so it doesn't kill the device Plugins?

Real bug, thanks for catching it. Fixed in 94c96d9:

- Startup: plugins with `DeviceNamespace` set spawn unconditionally, never gated on profile references.
- Deactivation sweep: excludes the set of UUIDs in `DEVICE_NAMESPACES` from removal candidates.

Device plugins serve hardware rather than per-profile actions, so their lifecycle shouldn't be tied to profile contents at all. I don't have a non-Elgato device on my rig, so the fix is scoped-and-reviewed rather than end-to-end tested — if you run that setup, a confirmation would be welcome.

Terrorwolf01:
Ok. As said, last time I checked, the official Software had all plugins running. This was about half a year or more ago tho.

I only own Elgato Devices myself, so can't test it unfortunately. Was just because as you said, it serves Hardware instead of Actions and needs to run all the time. How are you distinguishing Device Plugins from Action Plugins and how are you handling Plugins with both?

KatsuJinCode:
Mechanism: if `manifest.device_namespace.is_some()`, it's a Device Plugin. At runtime that's equivalent to "UUID is a value in the `DEVICE_NAMESPACES` map" since the map gets populated at plugin registration for any plugin with a non-empty `DeviceNamespace`.

Plugins that declare *both* (`DeviceNamespace` set, and `Actions[]` populated) are treated as Device Plugins for lifecycle purposes — spawn unconditionally at startup, never swept. The actions still function normally; the subprocess is just already up for device-serving reasons, so profile references become irrelevant to its lifecycle. Treating such a plugin as an action plugin would mean the attached device loses its host whenever the user switched to a profile with no actions from that plugin, which is the bug this PR already fixed for the pure-device case.

nekename:
I'm certain ESD spawns all plugins at startup and keeps them running. Not doing that would break assumptions around application monitoring and device connection events, deep linking, and so on. If anything, this creates more overhead as plugins have to reconnect both to OpenDeck and their own services frequently.

PR #325, opened 2026-05-09
Sleep when computer is locked support

### LINKED ISSUE #265: Disable devices while the computer screen is locked
**Preflight checklist**
- [x] I understand that if this issue is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [x] I understand that requests for the creation of new plugins or implementation of new actions should be opened as Discussions, not Issues.
- [x] I reasonably believe that the time required to implement this feature is proportional to the value it will provide to users.

---
Flow:
- User locks their machine
- Deck locks and/or turns "off"
- User unlocks
- Deck unlocks

StreamController has the option to lock your deck when one of their supported DEs are locked. However, for OpenDeck cross-platform needs to be solved.

Documenting this for discussion and gauging interest

D00Movenok:
### Preflight checklist

---

Resolve #265 

### Summary

- Add an optional setting to put Stream Deck devices to sleep when the computer is locked: `Sleep when computer is locked`.
- Track sleep reasons separately so `idle sleep`, `manual sleep`, and `computer-lock sleep` do not accidentally wake each other.

### Behavior

When `Sleep when computer is locked` is enabled, OpenDeck puts connected devices into sleep mode on screen lock and wakes them after screen unlock. Device inputs are intentionally swallowed while the computer is locked, so plugins do not receive key/dial events before unlock. If a device is (re)connected while the computer is still locked, it is immediately put back into sleep mode after registration.

Sleep Reason Effective Hierarchy:
1. `SLEEP_REASON_COMPUTER_LOCKED`
   Blocks device input and prevents commands from dispatching to plugins. Cleared only by screen unlock or disabling the setting.
2. `SLEEP_REASON_MANUAL`
   Set by `--sleep-device`. Cleared only by explicit wake command, currently `--wake-device`.
3. `SLEEP_REASON_IDLE`
   Set by inactivity timeout. Cleared by device activity, with the first activity swallowed so it only wakes the device.

Maybe it is better to merge `SLEEP_REASON_MANUAL` and `SLEEP_REASON_IDLE`. Feel free to do it/ask me for it.

### Tests

Tested it on my `Linux` system (Ubuntu 26.04) and on `Windows 10` VM. But I do not have an apple computers with `MAC OS`, so it will be better, if you (or someone else) test it on them. 

Feel free to ask questions :)

⚠️ ATTENTION ⚠️
I have added [this](https://github.com/pewsheen/psp) dependency to Cargo.toml, it is better to fork and use your fork instead of original repo. Also I has some additional ideas to that library, that can improve different systems coverage, so if you will fork it - I will create another PR.


src-tauri/src/elgato.rs
```diff
@@ -138,8 +141,6 @@ async fn init(device: AsyncStreamDeck, device_id: String) {
 	.await
 	.unwrap();
 
-	let reader = device.get_reader();
-	ELGATO_DEVICES.write().await.insert(device_id.clone(), device);
```
nekename: What is the reason for the changes in this file?

src-tauri/src/device_sleep.rs
```diff
@@ -1,16 +1,23 @@
 use std::sync::LazyLock;
-use std::sync::atomic::{AtomicU16, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU16, Ordering};
 use std::time::{Duration, Instant};
 
 use dashmap::DashMap;
 
+const SLEEP_REASON_IDLE: u8 = 1;
+const SLEEP_REASON_COMPUTER_LOCKED: u8 = 1 << 1;
+const SLEEP_REASON_MANUAL: u8 = 1 << 2;
```
nekename: Why is this not a standard Rust enum?

nekename (commented):
Thank you so much for a well-written PR! This is very refreshing to see.

The thing is, I would have implemented this myself a long time ago, except for one main blocker, that being the sheer outdatedness of `psp` and its dependencies. If you look at the changes to Cargo.lock here you'll see that older versions of many dependencies have been pulled in.

I believe my previous line was that I would be happy to accept this feature once someone has put in the work to update `psp`'s dependencies. The changes here, though, look very good.

src-tauri/src/elgato.rs
D00Movenok: On new device connection `register_device()` calls `apply_initial_device_sleep()` that calls `set_device_brightness()` that uses `ELGATO_DEVICES` (for elgato devices). So, without this patch - if new device is (re)connected to USB of logged-out PC, it will try to set brightness 0 (sleep mode) and can't find device in `ELGATO_DEVICES`. So I have moved adding to `ELGATO_DEVICES` before calling `register_device()`

src-tauri/src/device_sleep.rs
D00Movenok: This is my rust coding skills (i am real newbie in rust) :D
Let me fix this.

D00Movenok: Researched this thing, enums in rust (like in other languages) are incremental, e.g. elem1 = 0, elem2 = 1, elem3 = 2, elem4 = 3, but i have used bitmask in this patch, so i have 3 variants that i can implement:

1) Enums with custom definitions and `bits()` function as helper:
```rust
#[derive(Clone, Copy)]
#[repr(u8)]
enum SleepReason {
	Awake = 0,
	Idle = 1,
	ComputerLocked = 1 << 1,
	Manual = 1 << 2,
}

impl SleepReason {
	fn bits(self) -> u8 {
		self as u8
	}
}

reasons | reason.bits()
```

2) `bitflags` crate (seems architecture-cleaner for me and project already has it imported):
```rust
bitflags::bitflags! {
	struct SleepReason: u8 {
		const IDLE = 1;
		const COMPUTER_LOCKED = 1 << 1;
		const MANUAL = 1 << 2;
	}
}

let mut reasons = SleepReason::empty();
reasons.insert(SleepReason::IDLE);
reasons.insert(SleepReason::COMPUTER_LOCKED);
```

3) Just use vars (as i already used).

IMHO it is better to implement 2 or use 3 as it is already implemented.

What would you suggest?

nekename: Hmm? Why do you need an integer representation at all? Just use an enum and its variants

D00Movenok: Because, for example, ComputerLocked and Idle may be in the same time: user can't remove Idle pressing buttons because computer is locked, and after that we unlock computer (remove ComputerLocked bit), but it still stay sleeping - we need to press any button to remove Idle.

Maybe i missed something, but we can't combine two enum values in one?

Or, as i think, we can just merge all that sleep reasons to single. Now all that reasons seems overengineered.

D00Movenok: Removed all sleep reasons, but locked computer still locks `note_activity()` and any pressings.

nekename: My bad. I randomly remembered this PR in the shower and was wondering what I was even thinking, if I was thinking at all. I think I just hadn't read the full changes and didn't see the bitwise operations, presuming that you just stored a list of sleep reasons and not a bitmask. Came here prepared to see myself getting embarrassed in the comments, but you're right, this is a cleaner solution anyway :)

nekename:
As said, just let me know when/if psp gets updated, and I'd be glad to merge this. If psp doesn't want to accept changes upstream we can think about maintaining a fork

D00Movenok:
Ok, I will try to update it a little bit later

Why psp update is so necessary? Is it just the size and dependency list that matters, or is there something else?

Terrorwolf01:
> Ok, I will try to update it a little bit later
> 
> Why psp update is so necessary? Is it just the size and dependency list that matters, or is there something else?

Many outdated Dependencies that then get pulled double in multiple versions with maybe security vulnerabilities and such.

nekename:
Like Terrorwolf said, security and binary size.

D00Movenok:
@nekename I have updated psp library (now using my fork https://github.com/D00Movenok/psp)

Pls, check: is it ok now or I need to do something else? If ok, I'll try to merge psp fork to original repo and fix merge conflicts there.

nekename:
I'm not familiar with the psp codebase so I'm not a fit reviewer, but it looks good. Not sure why you changed the Error type etc but there must be a good reason; anyway, I'm sure the maintainer will be able to address any concerns. Thanks

D00Movenok:
PSP pr https://github.com/pewsheen/psp/pull/2

D00Movenok:
@nekename seems ready to merge

nekename:
I will look at this soon!

nekename:
@D00Movenok I've just pushed some changes - could you let me know if it still works correctly on your machine and if the changes look correct? Same for @Terrorwolf01?

nekename (approved):

Terrorwolf01:
Just tested on my System (EndeavourOS) and at lock the Streamdecks turned black, but after login, they stayed black until a restart of OpenDeck.

D00Movenok:
On my system (ubuntu 26.04) works perfectly after update

Terrorwolf01:
> On my system (ubuntu 26.04) works perfectly after update

Probably with Gnome Desktop? I am using KDE (Wayland) in the newest Version.

D00Movenok:
Yeah, gnome with wayland

nekename:
I'm also KDE Wayland and I think it was working. But I'm not at my computer anymore

nekename:
I've retested this on my PC (Plasma 6.7 Wayland) and on my MacBook (Sequoia 15.7.7) and it works without issues, both when locking the computer when the device is already sleeping due to the idle timeout and when the device is not yet sleeping.

I did find a deadlock when updating the idle timeout setting, but that was my fault (I introduced it in the `changes` commit). That's fixed now. For good measure, I asked Gemini 3.1 Pro to do a thorough review to find any outstanding issues and it only came up with things I think are acceptable.

Gonna merge :tada:
If Terrorwolf's issue is still reproducible we can fix it in a follow-up commit.

Thank you @D00Movenok for implementing an often-requested feature so thoroughly and meticulously. The only issues I encountered in this PR were my fault in my changes or me being stupid and not having the right settings on xD

You're very welcome to submit more PRs in the future should you wish to do so.

nekename:
Follow-up commit b9e47358a79fad8dcd70d29edbb494a3fb771c3d implements the `systemDidWakeUp` event using psp's Resume event

Nubinator:
I am on cachyos KDE Plasma Wayland 6.7 using the af29a43 commit and still face the issue with the streamdeck not waking up after locking the desktop. It happens regardless of whether the streamdeck was awake or asleep before locking the desktop

nekename:
Just tested on my machine again and I am indeed getting the issue as well :(

I'll see if I can get it fixed today (unless @D00Movenok gets to it first ;)) but otherwise it'll definitely be fixed before this is released.

D00Movenok:
Will take a look on nearest weekends

nekename:
Just tested with some `println` debugging and OpenDeck is no longer receiving the `ScreenUnlocked` event from psp at all. Maybe this is caused by a KDE update? I certainly wasn't experiencing this earlier but now it's reproducible - perhaps Plasma 6.7.

If so the only thing we may be able to do in OpenDeck is add a tooltip to the setting with a note, but it would be unfortunate...

nekename:
I'm receiving `ScreenLocked`, `Suspend` and `Resume` events but not `ScreenUnlocked`

nekename:
I did some further testing and it looks like running `loginctl unlock-session` allows devices to wake back up. So it's definitely not a bug in OpenDeck's logic, but perhaps somewhere in psp or otherwise in KDE. 

This means it could be a bug that wasn't only introduced in Plasma 6.7 because I was using KDE Connect to run `loginctl unlock-session` to test unlocking before (to avoid entering my password over and over), so that might explain why it was working for me earlier.

Terrorwolf01:
Also a side note: If you run into the Streamdeck staying black, you can remove the checkmark in Opendeck for this setting and then it "wakes up" again.

nekename:
I take it back. `gdbus monitor -y -d org.freedesktop.login1 | grep LockedHint` shows KDE sending the right events. Let me try to fix this in psp

nekename:
@D00Movenok I've opened https://github.com/pewsheen/psp/pull/3 which I think should fix the issue hopefully

Nubinator:
It seems to be working perfectly in 0ca1c3f on my system, Thanks!

nekename:
Excellent! v2.13.1 is coming out now

PR #339, opened 2026-05-22
fix: use a dedicated spawner thread so that plugin processes don't die due to PR_SET_PDEATHSIG when the parent Tokio worker exits

### LINKED ISSUE #329: Plugins processes die when the parent Tokio worker exits due to PR_SET_PDEATHSIG
### Preflight checklist

  **If you remove this checklist, this issue will be closed without explanation.**

  - [x] I understand that if this issue is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38, and that it should be raised in the repository of the plugin
  that provides support for the hardware.
  - [x] I will provide the OpenDeck log file and the log files of plugins involved, where applicable.
  - [x] I will provide my operating system information and method of installation, as well as any steps I have taken to troubleshoot the issue.
  - [x] I certify that I am using the latest available version of OpenDeck, and that I will provide enough information to reproduce the issue.

NOTE: This bug report has been co-authored by me, and Claude Code. I (the human) have spent, to the best of my abilities, the entire process supervising and rereading all text, descriptions and references to ensure that the quality is suitable for a bug report. I have also tried the suggested implementation and happy to provide a PR should that be required/desired.

  ---

  ## Summary

  On Linux, `attach_parent_death_signal` ([`src-tauri/src/plugins/mod.rs#L50-L61` @ `d6df914`](https://github.com/nekename/OpenDeck/blob/d6df914/src-tauri/src/plugins/mod.rs#L50-L61)) installs `PR_SET_PDEATHSIG =
  SIGTERM` on every native (and Node, and Wine) plugin child. This signal is delivered when the **specific OS thread inside OpenDeck that called `fork()` retires**, not when OpenDeck itself exits. Per `PR_SET_PDEATHSIG(2const)`:

  > The parent-death signal is sent upon subsequent termination of the **parent thread** and also upon termination of each subreaper process …

  and (CAVEATS, same page):

  > The "parent" in this case is considered to be the thread that created this process. In other words, the signal will be sent when that thread terminates (via, for example, `pthread_exit(3)`), rather than after all of the threads in the parent process terminate.

  Kernel-source corroboration: `forget_original_parent()` in `kernel/exit.c` dispatches `pdeath_signal` via `group_send_sig_info(... SEND_SIG_NOINFO ...)`, which is why our captured `siginfo_t` shows `si_code = SI_USER` and `si_pid = <opendeck PID>` rather than the `SI_KERNEL` value one might expect for a kernel-originated signal.

  Because `initialise_plugin` runs as a `tokio::spawn`'d task inside the multi-threaded tokio runtime ([`plugins/mod.rs#L403-L404` @ `d6df914`](https://github.com/nekename/OpenDeck/blob/d6df914/src-tauri/src/plugins/mod.rs#L403-L404)), the `fork()` call happens on an arbitrary tokio worker thread. Tokio workers are nominally long-lived, but in practice we observe them retiring (or being rotated) while OpenDeck itself continues running, which fires SIGTERM at the plugin child even though no legitimate teardown is in progress. The plugin's default disposition for SIGTERM is termination, so the plugin disappears without a trace: no panic, no clean WebSocket shutdown, no coredump.

  The "legitimate" plugin-termination path in `deactivate_plugin` ([`plugins/mod.rs#L378` @ `d6df914`](https://github.com/nekename/OpenDeck/blob/d6df914/src-tauri/src/plugins/mod.rs#L378)) uses `Child::kill()` which sends SIGKILL — so removing or reworking `PR_SET_PDEATHSIG` does not weaken explicit deactivation. The pdeathsig mechanism is purely a "if OpenDeck crashes, clean up orphan plugin processes" safety net, but with tokio's threading model it routinely fires *during normal operation*.

  ## How I encountered this

  I originally hit this bug while developing a native Rust replacement for the (abandoned) `com.baptiewright.nanoleaf` plugin. The original was a WebView/JS plugin, but OpenDeck's webview enforces CORS and the Nanoleaf REST API at `:16021` returns no CORS headers, so cross-origin XHR from the plugin webview is impossible — every request returns `status: 0` immediately. I rewrote the plugin as a native Rust binary against the `openaction` crate (which sidesteps the browser entirely). The native plugin worked, but died at random intervals — sometimes within seconds, sometimes after minutes of normal use — with no error, no panic, no log line. There was nothing in the per-plugin log file because OpenDeck truncates it on every respawn (`File::create`), so the death itself left no trace.

  After adding signal handlers and a heartbeat task, I confirmed the killer was SIGTERM. After adding a `sigaction(2)` handler that captures `siginfo_t`, I identified the source as `PR_SET_PDEATHSIG`. To rule out anything Nanoleaf-specific I reproduced the same death mode in the unmodified upstream `OpenActionPlugins/mpris` plugin, which is what the captured evidence below comes from.

  ## Reproduction

  1. Install any native Rust plugin built against `openaction = "2.x"` that does not explicitly handle SIGTERM. Confirmed reproducers:
     - `OpenActionPlugins/mpris` (upstream) — dies on Play/Pause around an album-art metadata change, but also at random idle moments.
     - A from-scratch Rust plugin (the Nanoleaf rewrite described above) — dies at random intervals during normal use.
     - `OpenActionPlugins/system`, `counter`, `discord`, `desktopentry` — all vulnerable in principle (none install a SIGTERM handler).
     - Webview-based plugins are immune (they're a Tauri `WebviewWindow`, not a child process).
  2. Run OpenDeck normally and use the plugin for anywhere from seconds to several minutes.
  3. The plugin process disappears silently. OpenDeck's UI may still display its keys, but presses no longer have any effect because no process is on the other end of the WebSocket.

  The timing is non-deterministic: short-lived plugins (e.g. counter) can persist; long-running plugins under any load eventually die.

  ## Root cause analysis

  I forked `OpenActionPlugins/mpris` and added a `sigaction(2)` handler that captures `siginfo_t` to atomics, logs `si_code` / `si_pid` / `si_uid` (and resolves `si_pid` to `comm`/`exe` via `/proc`), then ignores the signal and lets the process continue. After about 15 minutes of normal use, the plugin captured:

  ```
  11:56:04 [WARN] SIGTERM #1 caught — code=0 (SI_USER (kill(2) from another process))
    sender_pid=4088394 sender_uid=1000
    sender={pid=4088394 comm="opendeck" exe="/usr/bin/opendeck"}
    self={state=S (sleeping) ppid_raw=4088394 ppid=pid=4088394 comm="opendeck" exe="/usr/bin/opendeck" tracer=0}
    — IGNORING
  ```

  `code=SI_USER` and `sender_pid=<opendeck PID>` *look* like an explicit `kill(2)` from OpenDeck — but OpenDeck's source has no such call (only `child.kill()` = SIGKILL, and the `PR_SET_PDEATHSIG` setup). The kernel uses `SEND_SIG_NOINFO` when dispatching pdeathsig (`kernel/signal.c`), which fills `si_code = SI_USER` and `si_pid = task_tgid_nr_ns(current)` — i.e. the parent process's PID. At the siginfo level pdeathsig is therefore **indistinguishable from a real `kill()` from the parent**.

  Combined with the fact that OpenDeck was still running (our `ppid` is still its PID, not 1/init), this rules out every other plausible source:

  - Not OOM killer (would be SIGKILL and would coredump).
  - Not systemd-oomd (`journalctl -u systemd-oomd` empty for the window; pressure metrics nominal).
  - Not a Rust panic (panic hook installed, no record).
  - Not a clean WebSocket close (the openaction crate's final-return log line was never written).
  - Not SIGKILL (catchable by our handler, was caught).

  The only remaining source consistent with the captured siginfo is `PR_SET_PDEATHSIG` firing because a tokio worker thread inside OpenDeck retired.

  ## Affected plugins

  Every native Linux plugin built with `openaction` that does not explicitly install a SIGTERM handler. None of the five plugins in `OpenActionPlugins/*` (mpris, system, counter, discord, desktopentry) handle SIGTERM. Node plugins are vulnerable too in principle (same `attach_parent_death_signal` path applies to the `node` invocation at [`plugins/mod.rs#L259` @ 
  `d6df914`](https://github.com/nekename/OpenDeck/blob/d6df914/src-tauri/src/plugins/mod.rs#L259)); Node's default SIGTERM disposition is identical to Rust's, so they die the same way unless authors happen to install `process.on('SIGTERM', …)`.

  ## Suggested fixes

  The goal is to preserve the *intent* of the current code (clean up orphan plugins if OpenDeck itself dies) while not firing SIGTERM during normal operation. In order of preference:

  1. **Dedicate a single named OS thread for `Command::spawn()` and keep it alive for the program's lifetime.** Then `PR_SET_PDEATHSIG` semantics become "the spawner thread dies = OpenDeck is exiting", which is the intended behavior. A `std::thread::spawn` named `plugin-spawner` that reads commands off an `mpsc` channel and calls `Command::spawn()` for each is sufficient — ~30–50 lines of code, no behavior change for plugin authors.

  2. **Remove `attach_parent_death_signal` entirely.** Trade-off: if OpenDeck itself crashes (SIGSEGV / SIGABRT / etc.), plugin processes become orphans of init and continue running until they decide to exit. They can't do anything useful (no WS peer) but they consume memory until manually killed or the user logs out. For users who restart OpenDeck after a crash this manifests as duplicate plugin processes. Acceptable as a stopgap; not ideal as a permanent answer.

  3. **Document the SIGTERM requirement in `AGENTS.md` and require plugins to install an explicit handler.** Pushes the work onto every downstream author and silently breaks every plugin written without that knowledge. The five official plugins all need patches under this option.

  I'd suggest **(1)** as the actual fix and ship it sooner rather than later — it preserves cleanup-on-crash, eliminates the false-positive kills, and requires no changes to any plugin or to the `openaction` crate.

  ## Environment

  - OS: Fedora Linux 43 (Workstation Edition)
  - Kernel: 6.19.14-200.fc43.x86_64
  - OpenDeck: 2.12.0-1.x86_64 (Fedora RPM)
  - Reproducer plugins: `OpenActionPlugins/mpris` (upstream); a from-scratch native Rust plugin I'm developing (Nanoleaf controller rewrite).
  - Diagnostic source for the captured siginfo above: available on request — happy to publish the diagnostic fork of MPRIS if useful for verification.

nekename:
<details>

<summary>Preflight checklist</summary>

</details>

---

Fixes #329

### DIFF
--- a/src-tauri/src/plugins/mod.rs
+++ b/src-tauri/src/plugins/mod.rs
@@ -9,7 +9,7 @@ use crate::store::get_settings;
 
 use std::collections::HashMap;
 use std::process::{Child, Command, Stdio};
-use std::sync::LazyLock;
+use std::sync::{LazyLock, mpsc};
 use std::{fs, path};
 
 use tauri::{AppHandle, Manager};
@@ -20,6 +20,12 @@ use tokio::net::{TcpListener, TcpStream};
 use anyhow::anyhow;
 use log::{error, warn};
 use tokio::sync::{Mutex, RwLock};
+
+pub enum PluginChildType {
+	Wine,
+	Native,
+	Node,
+}
 
 enum PluginInstance {
 	Webview,
@@ -60,11 +66,14 @@ fn attach_parent_death_signal(command: &mut Command) {
 	}
 }
 
-/// Initialise a plugin from a given directory.
-pub async fn initialise_plugin(path: &path::Path) -> anyhow::Result<()> {
-	let plugin_uuid = path.file_name().unwrap().to_str().unwrap();
+pub type SpawnRequest = Box<dyn FnOnce() -> Result<(String, PluginChildType, Command), anyhow::Error> + Send>;
 
-	let mut manifest = manifest::read_manifest(path)?;
+/// Initialise a plugin from a given directory.
+pub async fn initialise_plugin(path: path::PathBuf, spawner_tx: mpsc::Sender<SpawnRequest>) -> anyhow::Result<()> {
+	let plugin_uuid = path.file_name().unwrap().to_str().unwrap().to_owned();
+	let plugin_uuid_2 = plugin_uuid.clone();
+
+	let mut manifest = manifest::read_manifest(&path)?;
 
 	if let Some(icon) = manifest.category_icon {
 		let category_icon_path = path.join(icon);
@@ -181,8 +190,15 @@ pub async fn initialise_plugin(path: &path::Path) -> anyhow::Result<()> {
 	}
 
 	let code_path = code_path.unwrap();
-	let port_string = PORT_BASE.to_string();
-	let args = ["-port", port_string.as_str(), "-pluginUUID", plugin_uuid, "-registerEvent", "registerPlugin", "-info"];
+	let args = [
+		"-port".to_owned(),
+		PORT_BASE.to_string(),
+		"-pluginUUID".to_owned(),
+		plugin_uuid.clone(),
+		"-registerEvent".to_owned(),
+		"registerPlugin".to_owned(),
+		"-info".to_owned(),
+	];
 
 	if code_path.to_lowercase().ends_with(".html") || code_path.to_lowercase().ends_with(".htm") || code_path.to_lowercase().ends_with(".xhtml") {
 		let url = format!("http://localhost:{}/", *PORT_BASE + 2) + path.join(code_path).to_str().unwrap();
@@ -215,7 +231,7 @@ pub async fn initialise_plugin(path: &path::Path) -> anyhow::Result<()> {
 			info = serde_json::to_string(&info)?
 		))?;
 
-		INSTANCES.lock().await.insert(plugin_uuid.to_owned(), PluginInstance::Webview);
+		INSTANCES.lock().await.insert(plugin_uuid, PluginInstance::Webview);
 	} else if code_path.to_lowercase().ends_with(".js") || code_path.to_lowercase().ends_with(".mjs") || code_path.to_lowercase().ends_with(".cjs") {
 		// Check for Node.js installation and version in one go.
 		let command = if is_flatpak() { "flatpak-spawn" } else { "node" };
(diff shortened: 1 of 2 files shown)


nekename:
@cstromblad @Terrorwolf01 please test :)

@FazalAAli @ambiso code review maybe? I'm a little rusty

ambiso:
I'm guessing the idea with the original was that if OpenDeck crashes then it will tear down all processes with it so we can do a clean restart. 

Does this separate spawner thread also die in case OpenDeck crashes?
If not I'd say this is missing the point a bit and we might as well remove the PDEATHSIG.

nekename:
> Does this separate spawner thread also die in case OpenDeck crashes?

I should hope so; it's a part of the OpenDeck process, not a separate process.

nekename:
@cstromblad I'm going to merge this later today, but if you get a chance to confirm that this resolves the issue in the meantime, that would be great.

cstromblad:
Hey,

Apologies for late reply, but I'm currently out travelling and have had limited time and access to my computer.

I'm hopefully back tomorrow and can then try the fix!

/ Christoffer

src-tauri/src/plugins/mod.rs
```diff
@@ -60,11 +66,14 @@ fn attach_parent_death_signal(command: &mut Command) {
...
 }
 
+pub type SpawnRequest = Box<dyn FnOnce() -> Result<(String, PluginChildType, Command), anyhow::Error> + Send>;
+
 /// Initialise a plugin from a given directory.
-pub async fn initialise_plugin(path: &path::Path) -> anyhow::Result<()> {
-	let plugin_uuid = path.file_name().unwrap().to_str().unwrap();
+pub async fn initialise_plugin(path: path::PathBuf, spawner_tx: mpsc::Sender<SpawnRequest>) -> anyhow::Result<()> {
+	let plugin_uuid = path.file_name().unwrap().to_str().unwrap().to_owned();
+	let plugin_uuid_2 = plugin_uuid.clone();
```
FazalAAli, line 74: nit: `plugin_uuid_2` is non descriptive, would recommend something like monitoring_uuid so it's more readable.

src-tauri/src/plugins/mod.rs
```diff
@@ -280,71 +283,62 @@ pub async fn initialise_plugin(path: &path::Path) -> anyhow::Result<()> {
...
 	}
 
 	if let Some(applications) = manifest.applications_to_monitor
 		&& let Some(applications) = applications.get(platform)
 	{
-		crate::application_watcher::start_monitoring(plugin_uuid, applications).await;
+		crate::application_watcher::start_monitoring(&plugin_uuid_2, applications).await;
 	}
 
 	Ok(())
```
FazalAAli, line 344: If you have scheduled a plugin to start, and the user immediately clicks remove, there is a chance that it hasn't been added to INSTANCES yet so remove_plugin will not be able to do it's job. Probably not an issue for most users but users with many plugins might see this behavior.

nekename: (We decided on Discord that since it would surface an error to the user, it's clear enough that the plugin wasn't removed correctly, meaning that it's clear enough that they should attempt removal again)

src-tauri/src/plugins/mod.rs
nekename: I think it's OK; monitoring_uuid would suggest it's a different UUID, but this is just a clone to work around ownership rules. A better name would be in order but I don't really feel like changing it now.

nekename:
@cstromblad do let us know if this has fixed your issue, since no one else was even able to reproduce it.

PR #341, opened 2026-05-23
fix: resolve race condition causing ENOENT error caused by concurrent accesses to the Settings store

nekename:
<details>

<summary>Preflight checklist</summary>

</details>

---

`Store::new()` cleans up the temp and backup files used to atomically save store content, which is not an issue for profile and device stores as they are only ever instantiated once, and never concurrently accessed (due to being accessed through RwLocks).

However, as the settings store (i.e. the store used for app settings) is accessed by multiple parts of the application without synchronisation and is instantiated multiple times, when the temp and backup files were cleaned up by creating a new instance, this could interrupt a save operation in progress by another instance.

This PR introduces a new method, `Store::new_concurrent()`, which skips the cleanup of the temp and backup files, so that Store can be instantiated multiple times without interfering with each other (the save operation is already managed by file locks).

Therefore, the `No such file or directory (os error 2)` error incurred when saving the application settings is resolved, as the temporary file is now never removed in the middle of a save operation.

Hopefully that makes sense!

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -119,7 +119,7 @@ async fn main() {
 				let _ = std::fs::rename(old, app.path().app_config_dir().unwrap());
 			}
 
-			let mut settings = store::get_settings()?;
+			let mut settings = store::get_settings();
 			use std::cmp::Ordering;
 			use tauri_plugin_dialog::{DialogExt, MessageDialogKind};
 			let current_version = semver::Version::parse(built_info::PKG_VERSION)?;
@@ -359,7 +359,7 @@ If you have already donated, thank you so much for your support!"#,
 				return;
 			}
 			if let WindowEvent::CloseRequested { api, .. } = event {
-				if let Ok(true) = store::get_settings().map(|store| store.value.background) {
+				if store::get_settings().value.background {
 					let _ = hide_window(window.app_handle());
 					api.prevent_close();
 				} else {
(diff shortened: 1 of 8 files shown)


src-tauri/src/main.rs
```diff
@@ -119,7 +119,7 @@ async fn main() {
 				let _ = std::fs::rename(old, app.path().app_config_dir().unwrap());
 			}
 
-			let mut settings = store::get_settings()?;
```
clxxiii, line 122: This should probably be `crate::settings::SETTINGS_MUT.blocking_lock()` so all mutations are running through the same lock

clxxiii (commented):
Looks good to me.
 It's probably better to use a [RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) for this use case to make sure both reads *and* writes are thread-safe. I imagine the OS handles most of the edge cases I can think of, but if you're still running into concurrency issues after this patch I'd recommend looking into that (unless there was a reason you explicitly didn't choose a RwLock).

nekename:
> Looks good to me. It's probably better to use a [RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) for this use case to make sure both reads _and_ writes are thread-safe. I imagine the OS handles most of the edge cases I can think of, but if you're still running into concurrency issues after this patch I'd recommend looking into that (unless there was a reason you explicitly didn't choose a RwLock).

Bildschirmfoto_20260525_172603

src-tauri/src/main.rs
nekename: I just tried to implement this change, but this setup hook is in a sync context but within the async runtime, meaning I can't .await on .lock(), but neither can I .blocking_lock() because it is technically within the Tokio runtime.

I could use `futures::executor::block_on` here, but do you think it's worth it for the extremely small risk that the update message is being shown at the same time as the frontend attempting to write the settings?

nekename: I'm going to take the calculated risk here and avoid blocking on the async runtime if not necessary.

PR #349, opened 2026-06-04
Adds support for StreamDeck + XL

CodeMartian:
### Preflight checklist

**If you remove this checklist, this pull request will be closed without explanation.**

- [X] I understand that if this pull request is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [X] I certify the compliance of this pull request with the [Jellyfin LLM/"AI" Development Policy](https://web.archive.org/web/20260414131310/https://jellyfin.org/docs/general/contributing/llm-policies/) adopted by the OpenDeck project.
- [X] I thoroughly understand the changes that I am proposing in this pull request, and I will be able to respond to questions and review feedback.
- [X] I reasonably believe that the changes that I am proposing are of production quality, or I am otherwise opening this pull request as a draft.
- [X] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.
- [X] I have ensured that I have run the appropriate formatter on my changes and that my code produces no linter violations.
- [X] I will keep "Allow edits from maintainers" enabled for this pull request.

---

PR #351, opened 2026-06-09
feat: add Stream Deck + XL device support and improve key layout

nschurmann:
### Preflight checklist

- [X] I reasonably believe that the changes that I am proposing are of production quality, or I am otherwise opening this pull request as a draft.

---

Changes in this PR:

1. Add Stream Deck + XL device recognition (bada6d9)
- Added support for identifying Stream Deck Plus XL devices in the device detection system
- Why: The original code didn't recognize the XL variant, so users with this device couldn't use it with OpenDeck
2. Use actual display instead of hardcoded values (da37186)
- Replaced hardcoded display dimensions with dynamic values read from the actual device
- Why: Hardcoded values didn't match the XL's display specs, causing incorrect rendering and layout issues
3. Add tests to display helpers (a004d43)
- Added unit tests for the display helper functions
- Why: To ensure the display logic works correctly and prevent regressions in future changes
4. Evenly distribute the knobs (3a9a83d)
- Adjusted the layout to evenly space encoder knobs on devices with multiple encoders
- Why: The previous layout had uneven spacing, making the UI look misaligned on devices like the Stream Deck Plus XL
5. Centering content supports Stream Deck + XL (d57b800)
- Updated the content centering logic to properly handle the XL device layout
- Why: The XL device has a different key arrangement that required adjustments to the centering algorithm
screenshot-2026-06-09_14-09-37

If any changes are need happy to update the PR.


nekename:
Thanks. I presume this supersedes #349. Could you sign your commits please?

src/components/DeviceView.svelte
```diff
@@ -87,6 +87,7 @@
 	];
 	$: encoderRowIndex = device.rows;
 	$: touchpointRowIndex = device.rows + (device.encoders > 0 ? 1 : 0);
+	$: deviceRowWidth = device.columns * 132;
```
nekename: Rename to keypadRowWidth maybe?

src/components/DeviceView.svelte
```diff
@@ -175,7 +176,7 @@
 	<span id="grid-description" class="sr-only">Use arrow keys to navigate between keys. Moving to a key will display its property inspector.</span>
 	<div
 		class="flex flex-col justify-center grow px-16 py-6 overflow-auto"
-		class:items-center={device.columns <= 8}
+		class:items-center={device.columns <= 9}
```
nekename, line 166: I don't remember what this code does. Why did you make this change?

src-tauri/src/elgato.rs
```diff
@@ -16,6 +16,32 @@ use tokio::sync::RwLock;
...
+	let encoder_count = kind.encoder_count() as usize;
+	let position = position as usize;
+	if position >= encoder_count {
+		return None;
+	}
+
+	if lcd_width >= lcd_height {
+		let segment_width = lcd_width / encoder_count;
+		Some(((position * segment_width) as u16, 0, segment_width as u32, lcd_height as u32))
+	} else {
```
nekename: Is this necessary? No Elgato devices have lcd_height > lcd_width. It can be added in the future if we encounter such a device.

nekename (commented):
@Terrorwolf01 could you check if this still displays on the SD+ please?

src/components/DeviceView.svelte
nschurmann: The elements are centered in the device view only if it has 8 columns or less, however the stream deck + XL has 9 columns.

src-tauri/src/elgato.rs
nschurmann: Not necessary, it was defensive programming. I'll take it out.

src/components/DeviceView.svelte
nekename: But I believe that that's to do with the fade-out and overflow logic that kicks in for devices wider than 8 columns. Shouldn't that behaviour be preserved for the SD+XL?

Terrorwolf01:
> @Terrorwolf01 could you check if this still displays on the SD+ please?

Seems to still work on the normal SD+.

src/components/DeviceView.svelte
nschurmann: At least for me the overflow logic is not working. I'm attaching screenshots on how it's behaving with device columns <= 8.

I'm using a Samung ultrawide 34 inch, first image you can see how the content aligns to the left.
screenshot-2026-06-11_11-53-58
Second image you can see how the content is being "squished" and the overflow logic is not being activated. 
screenshot-2026-06-11_11-55-36

However if I leave it at 9:
screenshot-2026-06-11_12-02-40

screenshot-2026-06-11_12-02-56
 
Your call, we can leave it at 9 or 8.

src/components/DeviceView.svelte
nschurmann: Sure!

nschurmann:
PR updated, what's left is to decide what to do with `class:items-center={device.columns <= 8|9}` rule @nekename

nekename:
@nschurmann The new changes all look good. I'm wondering if we should just have unconditional items-center on that element... I'll see if I can do some testing to determine the original reason for that class. Otherwise, I'm going to wait until https://github.com/OpenActionAPI/rust-elgato-streamdeck/issues/58 is resolved to merge this, so that this PR can update the crate to a version with the necessary fix.

src-tauri/src/elgato.rs
```diff
@@ -111,7 +139,7 @@ async fn init(device: AsyncStreamDeck, device_id: String) {
 		Kind::Mini | Kind::MiniMk2 | Kind::MiniDiscord | Kind::MiniMk2Module => 1,
 		Kind::Xl | Kind::XlV2 | Kind::XlV2Module => 2,
 		Kind::Pedal => 5,
-		Kind::Plus => 7,
+		Kind::Plus | Kind::PlusXl => 7,
```
nekename: One more thing: the PlusXL is not device type 7. https://docs.elgato.com/streamdeck/sdk/guides/devices#device-types

nekename:
Alright, I've tested the items-center behaviour with a 10x6 device.

### Without items-center

Bildschirmfoto_20260611_122637
Bildschirmfoto_20260611_122825
Bildschirmfoto_20260611_123413


As you can see, at the automatically-resized/minimum window size, perhaps it unnecessarily adds space on the left, but at least the entire grid is visible. When the window is maximised, the grid is wrongly aligned to the left.


### With items-center

Bildschirmfoto_20260611_123136
Bildschirmfoto_20260611_123205
Bildschirmfoto_20260611_123218

Whereas with items-center, the grid is clipped off on the left-hand side at the automatically-resized/minimum window size, but it correctly centres the content when maximised.

What we want is a mix of the two: no clipping the grid at the minimum window size, but the content centred when maximised.

Do you think you could find a solution that achieves both things?

nschurmann:
@nekename I think we could but I realized after some testing that the "squishing window" only happens in my setup with is omarchy with hyprland, so I wont be able to test it properly an the content doesn't overflow for me. Can we move this issue to a separate PR? I'll update the mapping for the device ID.

Terrorwolf01:
> @nekename I think we could but I realized after some testing that the "squishing window" only happens in my setup with is omarchy with hyprland, so I wont be able to test it properly an the content doesn't overflow for me. Can we move this issue to a separate PR? I'll update the mapping for the device ID.

For the Squishing you have, that shouldn't happen in the first place. The window size is limited to a minimum normally. Maybe a bug in combination with Hyprland.

nschurmann:
@Terrorwolf01 I agree, I tested this also in a fresh Tauri app with nothing in it and it squishes after a certain point. So I think it's my hyprland/wayland combination setup.

nekename:
Besides the window squishing, there is also the grid alignment issue, which needs fixing (+ the device type value).

nekename:
Just wanted to say that I'm only waiting on the elgato-streamdeck issue (and the device type) to merge this. I can fix the grid alignment myself, because that's my code. I'm going to aim to do 2.13.0 sooner rather than later and there won't be a 2.14 for a while while I enjoy my summer, and this wouldn't make sense for a 2.13.1, so if anyone has any updates, that would be great.

nekename:
I merged in main with the encoder layout support, fixed the device type and fixed not having the right udev rules

Thanks for the PR

nekename:
@nschurmann or @CodeMartian could you let me know if what I've merged actually works? I'm having doubts...

nekename:
It was not in fact correct, b80fed355d5f109c1a620b1aa09c3650331bda28 fixes it

Issue #59, opened 2025-02-26
Key press action does not repeat key until released


tlrib:
Hi! 

When configuring a Stream Deck Pedal button in OpenDeck to send [Key(RightArrow, Press)] on key down and [Key(RightArrow, Release)] on key up, the key does not repeat as expected when held down. Instead, it acts as a single key press, meaning the cursor (or any action mapped to the key) only moves once rather than continuously.

On a regular keyboard, holding the Right Arrow key causes the system to generate repeated key events automatically. However, with OpenDeck, this behavior does not occur.
Steps to Reproduce:

-  Configure a Stream Deck Pedal button in OpenDeck as follows: Key Down:

![Image](https://github.com/user-attachments/assets/0422b386-31f0-454c-95fb-b9eab905218e)

-  Press and hold the pedal.
-  Observe that the cursor (or mapped function) does not repeat like a normal keyboard key would.

**Expected Behavior:**

When the pedal is held, the assigned key should repeat just like a physical keyboard key does when held down.

**Actual Behavior:**

Only a single press event is registered, and no repeat occurs.

Is there a way to achieve this? I've been trying to look in the source code and didn't find a way. Maybe I'm doing something wrong.
Thank you!

ninjadev64:
Hmm, while I see where you're coming from with regards to the behaviour of a real keyboard, I think that press according to the library's definition is more of a "key down" event rather than simulating a real key press.

This functionality would have to be implemented upstream (`enigo-rs/enigo`) so I'm going to close this issue as an upstream issue.

Dessix:
I just ran into this myself- the problem seems to be that the enigo context is being dropped between down and up events. If you keep the context around, this will just work. Alternatively, as a hack, you could just set it to [not release held keys on drop](https://github.com/enigo-rs/enigo/blob/e5773b56a9383e30ac215a0bd40250bc30d7f54f/src/lib.rs#L454).
Otherwise, this tool is basically useless for push-to-talk scenarios.

PR #354, opened 2026-06-15
fix(macos): stop sleep/wake crash by pinning hidapi to a dedicated thread

kingb:
### Preflight checklist

- [] I understand that if this pull request is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [] I certify the compliance of this pull request with the [Jellyfin LLM/"AI" Development Policy](https://web.archive.org/web/20260414131310/https://jellyfin.org/docs/general/contributing/llm-policies/) adopted by the OpenDeck project.
- [] I thoroughly understand the changes that I am proposing in this pull request, and I will be able to respond to questions and review feedback.
- [] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.
- [] I have ensured that I have run the appropriate formatter on my changes and that my code produces no linter violations.
- [] I will keep "Allow edits from maintainers" enabled for this pull request.

---

## Summary

On macOS, OpenDeck reliably crashes a few seconds after the machine wakes from
sleep, with an `EXC_BREAKPOINT` inside CoreFoundation:

```
__CFCheckCFInfoPACSignature
  ← CFRunLoopAddSource
  ← IOHIDDeviceScheduleWithRunLoop
  ← hidapi hid_open / hid_enumerate
  ← elgato_streamdeck::AsyncStreamDeck::connect
```

### Root cause

hidapi's macOS backend binds its `IOHIDManager` to the run loop of the thread
that created the `HidApi`. Enumerating or opening a device from any *other*
thread schedules IOKit sources on a foreign run loop, which trips a
CoreFoundation pointer-authentication check and traps. This is most reliably
hit right after a sleep/wake, when the periodic device poll runs while the
device set is changing.

The current code stores a single `HidApi` in a `LazyLock<RwLock<Option<Arc<…>>>>`
and calls the async enumerate/connect wrappers from whatever tokio worker thread
happens to run `initialise_devices()`. elgato-streamdeck's async wrappers use
`block_in_place`, which does **not** pin work to one OS thread (confirmed even on
a single-worker runtime), so the `IOHIDManager` ends up driven from a foreign
thread.

### Fix

Create the `HidApi` once on a dedicated, long-lived OS thread (`opendeck-hid`)
and run *every* enumerate/open on that same thread via the **synchronous**
elgato-streamdeck API (which never calls `block_in_place`). Opened `StreamDeck`
handles are `Send`, so they're handed back over a channel and wrapped with
`AsyncStreamDeck::from(device)` for async use as before. `initialise_devices()`
now sends a scan request to that thread and awaits the opened devices.

No behavior change to device handling otherwise — same `sd-<serial>` IDs, same
`init()` path, same dedup against already-registered devices.

## Dependency note (why this needs an elgato-streamdeck release)

The fix wraps an already-opened sync `StreamDeck` for async use, which needs
`impl From<StreamDeck> for AsyncStreamDeck`. That was added upstream in
OpenActionAPI/rust-elgato-streamdeck#59 (merged), but is **not yet in a
published crates.io release** — the latest release predates the merge.

To keep this buildable today, `Cargo.toml` temporarily points
`elgato-streamdeck` at a fork branch carrying that commit on top of the 0.12.1
release (the version OpenDeck currently pins):

```toml
[patch.crates-io]
elgato-streamdeck = { git = "https://github.com/kingb/rust-elgato-streamdeck", branch = "feat/async-from-sync-0.12" }
```

**This `[patch.crates-io]` is a placeholder and should not merge as-is.** Once a
release of elgato-streamdeck containing #59 is published, the patch should be
dropped and the dependency bumped to that version. Opening this now to surface
the macOS fix and to motivate cutting that release. Happy to update the PR to a
plain version bump the moment a release is available.

(If the release lands as 0.13.x rather than a 0.12.x point release, the bump
will also need a one-line `Kind::PlusXl` arm added to the `device_type` match in
`elgato.rs`, since 0.13 introduced that device variant and the match is
exhaustive. Not included here because OpenDeck still pins 0.12.)

## Testing

- Reproduced the original crash: stock build, sleep the Mac, wake it, crash
  within ~10s. Symbolicated the report to the trace above.
- With the fix: confirmed the `opendeck-hid` thread is present, device
  enumeration/open run on it, and the app survives repeated sleep/wake cycles
  with the Stream Deck attached and re-enumerating.
- Normal hotplug (unplug/replug while running) still registers devices.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1306,8 +1306,7 @@ checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 [[package]]
 name = "elgato-streamdeck"
 version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e86db7c0f5bbffc406f4226ea5bf900bfd818bd8e4f93482bdc1ceeb17824ac"
+source = "git+https://github.com/kingb/rust-elgato-streamdeck?branch=feat%2Fasync-from-sync-0.12#cb2c983054370b8223cb7bdeb53eb721d0e95b2a"
 dependencies = [
  "hidapi",
  "image",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -61,3 +61,11 @@ windows-sys = "0.61"
 
 [features]
 custom-protocol = ["tauri/custom-protocol"]
+
+# elgato-streamdeck with an added `From<StreamDeck> for AsyncStreamDeck`, so
+# device enumeration/open can be pinned to a dedicated HID thread (see
+# elgato.rs). Fixes a macOS sleep/wake crash where hidapi's IOHIDManager was
+# driven from a foreign thread's run loop. Drop this patch once the upstream PR
+# lands in a release: https://github.com/OpenActionAPI/rust-elgato-streamdeck/pull/59
+[patch.crates-io]
+elgato-streamdeck = { git = "https://github.com/kingb/rust-elgato-streamdeck", branch = "feat/async-from-sync-0.12" }
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -1,20 +1,87 @@
 use crate::events::inbound;
 
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::collections::HashSet;
 use std::sync::LazyLock;
 
 use base64::Engine as _;
 use elgato_streamdeck::{
-	AsyncStreamDeck, DeviceStateUpdate,
+	AsyncStreamDeck, DeviceStateUpdate, StreamDeck,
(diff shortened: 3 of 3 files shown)


nekename:
You've removed the PR preflight checklist, so I can't know if you're following the policies it asks you about.

nekename:
Alright, you re-added the checklist, but it doesn't look like you've actually ticked any of the boxes.

nekename:
As it stands, this PR is in violation of section 1 of the General Guidelines of the LLM/AI policy, and will be closed as such.

kingb:
Understood. I got distracted by other work. My intent is to fully-test an Ubuntu laptop as well, and not just mac to better confirm I didn't break anyting and then I'll re-open and fill out the form

nekename:
I closed the PR because the PR description was fully AI-written, which suggests little to no human oversight was involved in developing the changeset.

PR #356, opened 2026-06-16
Add support for StreamDeck+ Strip Rendering

FrostyCoolSlug:
### Preflight checklist

- [✓] I understand that if this pull request is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [✓] I certify the compliance of this pull request with the [Jellyfin LLM/"AI" Development Policy](https://web.archive.org/web/20260414131310/https://jellyfin.org/docs/general/contributing/llm-policies/) adopted by the OpenDeck project.
- [✓] I thoroughly understand the changes that I am proposing in this pull request, and I will be able to respond to questions and review feedback.
- [✓] I reasonably believe that the changes that I am proposing are of production quality, or I am otherwise opening this pull request as a draft.
- [✓] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.
- [✓] I have ensured that I have run the appropriate formatter on my changes and that my code produces no linter violations.
- [✓] I will keep "Allow edits from maintainers" enabled for this pull request.

---

This is the initial pass on strip rendering, it's not currently rendering in the UI (we discussed this on discord, and it makes sense to handle that separately). but will render on-device. There's some small bits of code which pass information back and forth to the UI, although they're placeholders until the full work can begin there.

It currently leverages https://github.com/FrostyCoolSlug/streamdeck-strip-render to perform the drawing in rust to aid with performance (especially considering some actions may do frequent updates, and webkitgtk's performance under linux isn't really friendly to rapid complex canvas redrawing).


FrostyCoolSlug:
Sorry, force push needed to properly Verify the commits :)

src-tauri/src/events/inbound/states.rs
```diff
@@ -101,6 +109,146 @@ pub async fn set_image(mut event: ContextAndPayloadEvent<SetImagePayload>) -> Re
...
+		&& let Some(encoder) = &mut action.action.encoder
+	{
+		// We need to replace the existing parsed layout with the new one
+		encoder.layout = event.payload.layout.clone();
+
+		// Make sure the layout is a full path to the json file
+		if !encoder.layout.starts_with("$") {
+			let path = config_dir().join("plugins").join(&action.action.plugin);
+			encoder.layout = path.join(&encoder.layout).to_string_lossy().to_string();
+		}
```
nekename: Just in case, it might be nice to either sanitize &encoder.layout first, or canonicalise and prefix-check the joined path afterwards. I know it's crazy to even suggest this being exploited, but hypothetically, the contents of `/etc/shadow` showing up on the Stream Deck of a live streamer who only installed web-based plugins they thought couldn't read files would be bad... since this PR is so well written, we might as well have perfection.

src-tauri/src/elgato.rs
```diff
@@ -37,14 +41,67 @@ pub async fn update_image(context: &crate::shared::Context, image: Option<&str>)
...
 			if context.controller == "Encoder" {
-				device
-					.write_lcd(
-						(context.position as u16 * 200) + 64,
-						14,
-						&ImageRect::from_image_async(image::load_from_memory(&bytes)?.resize(72, 72, image::imageops::FilterType::Nearest))?,
-					)
-					.await?;
+				let locks = acquire_locks().await;
+				if let Some(action) = get_slot(context, &locks).await?
```
nekename: This is an `instance`, not an `action` - changing the name would also avoid the ambiguous `action.action` action going on below

src-tauri/src/shared.rs
```diff
@@ -218,10 +219,99 @@ pub struct Action {
...
+	#[serde(alias = "StackColor")]
+	pub stack_color: String,
+
+	#[serde(default, alias = "TriggerDescription")]
+	pub trigger_description: TriggerDescription,
+
+	#[serde_inline_default(String::new())]
+	pub background: String,
+
+	// TODO: Validation Pattern: "^(^(?![\\.]*[\\\\\\/]+).+\\.([Jj][Ss][Oo][Nn])$)|(\\$(X1|A0|A1|B1|B2|C1))$"
```
nekename: Is this planned for this PR, or even necessary?

src-tauri/src/shared.rs
```diff
@@ -357,3 +447,194 @@ pub static CATEGORIES: LazyLock<RwLock<HashMap<String, Category>>> = LazyLock::n
 	);
 	RwLock::new(hashmap)
 });
+
+// TODO: These probably can go somewhere else, I'm not sure where to put them :D
```
nekename: These should be `include_str!`'ed, maybe from `/static/encoder_layouts`?

src-tauri/Cargo.lock
```diff
@@ -2,6 +2,22 @@
 # It is not intended for manual editing.
 version = 4
```
nekename, line 4: Wow, this PR sure adds a lot of dependencies... any chance that could be trimmed down in the strip-renderer lib?

src-tauri/src/elgato.rs
```diff
@@ -37,14 +41,67 @@ pub async fn update_image(context: &crate::shared::Context, image: Option<&str>)
 		if let Some(image) = image {
 			let data = image.split_once(',').unwrap().1;
 			let bytes = base64::engine::general_purpose::STANDARD.decode(data)?;
+
 			if context.controller == "Encoder" {
```
nekename, line 88: Would it make sense for this to be a function called independently of update_image, or at the very least, a separate function called by update_image - just to reduce the number of nested blocks and the length of the function here? Even just cut-and-pasting this logic into a function above and subsituting it with a call to the new function would make this a lot more readable.

nekename (commented):
Overall, this is a very polished and well-written PR and was refreshing to review. I have a few minor thoughts and a hopeful request for a reduction in added dependencies, but this is looking good otherwise.

src-tauri/src/events/inbound/states.rs
FrostyCoolSlug: I understand the defensiveness, although this would only be a problem if `/etc/shadow` is JSON and follows the full strip schema (it would otherwise get immediately rejected either during the json parse, or after being passed to the lib), so arbitrary file loading wouldn't actually result in anything bad. With that said, I'll add a check.

src-tauri/src/shared.rs
FrostyCoolSlug: `load_encoder_layout` actually handles this, so the TODO isn't needed anymore, I'll remove this.

src-tauri/src/elgato.rs
FrostyCoolSlug: Will do, this started at about 8 lines long, and evolved, so I'll move it out.

src-tauri/Cargo.lock
FrostyCoolSlug: I took a look and removed a couple which aren't needed (loading images inside SVGs) but that only really cut it down by 7, unfortunately otherwise it's as tight as it can be. 16 of the dependencies are shared with upstream OpenDeck, but the other 36 are all related to image drawing, and more specifically, SVG rendering which encompasses about 70% of those added.

src-tauri/src/events/inbound/states.rs
FrostyCoolSlug: Resolved in 23e27fc, 7fe3279 and ff5b31b

src-tauri/src/elgato.rs
FrostyCoolSlug: Resolved in 5467141

src-tauri/src/shared.rs
FrostyCoolSlug: Resolved in 10b5382

src-tauri/src/shared.rs
FrostyCoolSlug: Yea, that's much cleaner, resolved in 5649ae9

src-tauri/src/elgato.rs
FrostyCoolSlug: Resolved in 9338f99, this was a good shout, and will actually make life easier when the UI side is handled as well, because we can call the same function to get the image :)

src-tauri/Cargo.lock
FrostyCoolSlug: I'm happy to flag this as resolved, because there really isn't much else I can do to improve this, but I'll leave it unless you can suggest anything I'm not thinking of?

nekename: Yeah, thanks for trimming it down a little

nekename (approved):

PR #357, opened 2026-06-17
feat: add system sleep/wake recovery for devices and webview plugins

GDWhisper:
## Summary

On Windows, system sleep/hibernation causes three categories of failure that leave
Stream Deck devices and plugins non-functional after resume:

1. **USB HID disconnection** — devices lose their HID handles and must be
   re-enumerated.
2. **WebSocket disconnection** — the plugin WebSocket server loses all active
   connections; neither the Elgato SDK nor the openaction SDK has built-in
   reconnection logic.
3. **Web Worker timer death** — webview plugins that use the Elgato SDK's
   `timers.js` have their `setTimeout`/`setInterval` overridden by a Web Worker.
   This Worker dies during sleep and does not resume, so even after WebSocket
   reconnection the plugin's timers (e.g. clock animations) remain frozen.

This PR adds a `system_sleep_watchdog` module that detects system sleep via
wall-clock monitoring and orchestrates a full recovery sequence on wake.

### New: `system_sleep_watchdog.rs`
- Polls wall-clock time at 1 Hz; detects sleep when elapsed > 3 s between polls.
- On wake: waits 2 s for USB stabilisation → re-enumerates devices → wakes idle
  devices → broadcasts `systemDidWakeUp` to all plugins → reloads webview plugins.

### New: `events/outbound/misc.rs`
- `system_did_wake_up()` sends a `systemDidWakeUp` JSON event to every connected
  plugin (native and webview), matching the Elgato SDK convention so plugins can
  optionally react to wake events.

### Modified: `plugins/mod.rs`
- `reload_webview_plugins()` — for each webview plugin:
  1. `location.reload()` to get a fresh JS environment and new Web Worker.
  2. Hidden-iframe trick to restore native `setInterval`/`setTimeout` (the new
     Worker after reload may also be unreliable on some systems).
  3. Re-evaluate the SDK connection init JS to create a fresh WebSocket and send
     `registerPlugin`.
- Avoids closing/recreating the Tauri window, since `window.close()` does not
  reliably destroy webviews on Windows (causes "label already exists" panics).

### Modified: `main.rs`
- Register `system_sleep_watchdog` module and call `init_watchdog()` on startup.

## Performance Impact

The overhead is negligible:

- **CPU**: The watchdog sleeps on a tokio interval and wakes once per second to
  compare two `SystemTime` values — effectively a single syscall per second. No
  busy-waiting, no polling of device handles or file descriptors. On wake, device
  re-enumeration and webview reload are one-shot operations that complete in a few
  seconds and then go idle.
- **Memory**: One persistent tokio task (~100 bytes of state). The hidden iframe
  created per webview plugin is a lightweight DOM element that persists but
  consumes no CPU and negligible heap.
- **Network**: No additional network traffic. WebSocket reconnection reuses the
  existing localhost port.

In short, the idle-state cost is one 1 Hz timer tick; all heavier work only
happens during the brief wake-recovery window.

## Tested

- Windows 11, AKP153 + onairclock webview plugin.
- Confirmed recovery after 68-minute sleep: USB device responsive, webview clock
  ticking within seconds of wake.

## Suggestion

It may be worth documenting the `systemDidWakeUp` event in the plugin development
guide so that plugin authors are aware they can subscribe to it for their own
sleep-recovery logic. The event follows the Elgato SDK naming convention and
carries no payload:

```json
{ "event": "systemDidWakeUp" }
```

Native plugins could use it to re-open USB handles; webview plugins could use it
to rebuild application state — though for webview plugins, OpenDeck already
handles full reconnection automatically.

## Notes

- The wall-clock approach is deliberately simple and cross-platform — it does not
  depend on platform-specific power-event APIs.
- The 3 s threshold is well above the 1 s poll interval even with scheduler noise,
  so false positives are extremely unlikely.

### DIFF
--- a/src-tauri/src/events/outbound/misc.rs
+++ b/src-tauri/src/events/outbound/misc.rs
@@ -1,0 +1,15 @@
+use super::send_to_all_plugins;
+
+use serde::Serialize;
+
+#[derive(Serialize)]
+struct SystemDidWakeUpEvent {
+	event: &'static str,
+}
+
+pub async fn system_did_wake_up() -> Result<(), anyhow::Error> {
+	send_to_all_plugins(&SystemDidWakeUpEvent {
+		event: "systemDidWakeUp",
+	})
+	.await
+}
--- a/src-tauri/src/events/outbound/mod.rs
+++ b/src-tauri/src/events/outbound/mod.rs
@@ -3,6 +3,7 @@ pub mod deep_link;
 pub mod devices;
 pub mod encoder;
 pub mod keypad;
+pub mod misc;
 pub mod property_inspector;
 pub mod settings;
 pub mod states;
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -8,6 +8,7 @@ mod events;
 mod plugins;
 mod shared;
 mod store;
+mod system_sleep_watchdog;
 mod zip_extract;
 
 mod built_info {
@@ -197,6 +198,7 @@ If you have already donated, thank you so much for your support!"#,
 			plugins::initialise_plugins();
 			application_watcher::init_application_watcher();
 			device_sleep::init_device_sleep();
+			system_sleep_watchdog::init_watchdog();
 
 			let label = IconMenuItemBuilder::with_id("label", PRODUCT_NAME)
 				.icon(app.default_window_icon().unwrap().clone())
--- a/src-tauri/src/plugins/mod.rs
+++ b/src-tauri/src/plugins/mod.rs
@@ -391,6 +391,146 @@ pub async fn deactivate_plugins() {
 	}
 }
 
+/// Reconnect all webview plugins after system wake.
+///
+/// Webview plugins lose both their WebSocket connection AND their Web Worker
+/// timers when the system enters sleep.  The Elgato SDK uses a Web Worker
+/// (`timers.js`) that overrides `setTimeout`/`setInterval`; this Worker dies
+/// during sleep and does not resume, so even if the WebSocket were reconnected
+/// the plugin's timers (e.g. clock animations) would remain frozen.
+///
+/// The fix is a two-step process:
+///   1. Reload the webview page (`location.reload()`) — this gives the plugin
+///      a fresh JavaScript environment with a new Web Worker and fresh timer
+///      overrides.
+///   2. Restore native browser timers via a hidden iframe and re-evaluate the
+///      connection init JavaScript — this creates a new WebSocket to the
+///      OpenDeck server and sends `registerPlugin`.
+///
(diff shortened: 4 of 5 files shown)


Terrorwolf01:
Could you check if https://github.com/nekename/OpenDeck/pull/325 already fixes your issues? Right now opendeck doesn't has any support yet for Sleep of the PC on any Platform. For such features, we try to have as much Cross-Plattform support instead of splitting everything up for each Software.

nekename:
Alright, this looks LLM-generated. Are you complying with the OpenDeck LLM policy? I can't tell, because you removed the preflight checklist that asks about your compliance.

nekename:
Hey, actually, I just realised that it says 'If you remove this checklist, this pull request will be closed without explanation.', so I am going to close this PR.

GDWhisper:
> Could you check if #325 already fixes your issues? Right now opendeck doesn't has any support yet for Sleep of the PC on any Platform. For such features, we try to have as much Cross-Plattform support instead of splitting everything up for each Software.

I think #325 is for the lock screen scenario. My issue is system sleep — devices and plugins stop working after wake. Tried it on my machine and it fixed the problem for me.

GDWhisper:
> Alright, this looks LLM-generated. Are you complying with the OpenDeck LLM policy? I can't tell, because you removed the preflight checklist that asks about your compliance.

Sorry, that was my oversight. I didn’t notice the OpenDeck LLM policy and the PR template earlier. I’ve gone through them just now. In fact, I know exactly what changes the LLM made, and I was the one who instructed how the PR should be written.

nekename:
Alright, I've reopened this. Please could you re-add the checklist and tick all the boxes and then I can take a look.

nekename:
But to be honest, this many Windows-specific changes including some things in the changeset that look pretty sketchy, as well as another PR already being open to add proper power state monitoring using the psp crate, this PR (which seems like just a bunch of hacky workarounds and no real solutions) is not going to be merged.

nekename:
@GDWhisper now that #325 is merged, you're welcome to give this another go using the proper power monitoring afforded by psp. Also, thank you for bringing my attention to the lack of the `systemDidWakeUp` event, which I've implemented in b9e47358a79fad8dcd70d29edbb494a3fb771c3d using the aforementioned psp features.

GDWhisper:
> @GDWhisper now that #325 is merged, you're welcome to give this another go using the proper power monitoring afforded by psp. Also, thank you for bringing my attention to the lack of the `systemDidWakeUp` event, which I've implemented in [b9e4735](https://github.com/nekename/OpenDeck/commit/b9e47358a79fad8dcd70d29edbb494a3fb771c3d) using the aforementioned psp features.

I tried psp on Windows, works fine. Definitely more reliable than polling.

PR #359, opened 2026-06-17
Explicit device layout

BrendanGrant:
### Preflight checklist

---

At present, OpenDeck devices can only describe their layout using a small set of fixed properties: button matrix rows and columns, encoder count, and touchpoint count. That works well for regular devices like the Stream Deck Mk.2, XL, and Neo, but it does not model irregular devices cleanly.

OpenDeck’s policy is that only Elgato hardware is officially supported in the core application, while support for other hardware vendors belongs in plugins. This PR does not add built-in support for any non-Elgato device. Instead, it extends the plugin registration model so plugins can describe the layout of the devices they already register.

This change adds an optional explicit layout model that lets a plugin provide the position, size, shape, and surface grouping for each control. The existing registration shape remains supported, so plugins that only provide rows, columns, encoders, and touchpoints continue to use the legacy renderer.

I have been testing this in two ways: with a fake-device plugin branch and with a real Loupedeck Live connected to a Raspberry Pi 3 running a build of OpenDeck based on this change. The Loupedeck plugin is here: [BrendanGrant/OpenDeck.Loupedeck](https://github.com/BrendanGrant/OpenDeck.Loupedeck). The fake-device plugin branch is here: [OpenDeck.FakeDevicesPlugin.Web fake_devices](https://github.com/BrendanGrant/OpenDeck.FakeDevicesPlugin.Web/tree/fake_devices). The installable fake plugin package can be downloaded here: [opendeck-fake-devices-web.sdPlugin](https://github.com/BrendanGrant/OpenDeck.FakeDevicesPlugin.Web/blob/fake_devices/output/opendeck-fake-devices-web.sdPlugin).

This is working well in my testing, but I am opening the PR as a draft to solicit feedback on the approach. In particular, I would like feedback on whether this explicit layout model is a good direction for plugin-provided devices, and whether the shape of the registration metadata feels appropriate for OpenDeck long term.


nekename:
Thank you for the submission and for the Loupedeck plugin. Re: Loupedeck, someone else just showed off a Loupedeck  plugin a day or two ago: https://github.com/Aib0t/opendeck-loupedeck - in case you were interested.

More flexible device layouts have often been requested so I'd be happy to take a look at this soon.

Terrorwolf01:
Just to clarify since I don't think I saw it, does this Support LCD Screens like they are on the SD+? And how about touch Points like on the SD Neo?

(Sorry if I just missed it, its late already)

BrendanGrant:
@nekename 

> Re other Loupedeck

Ha! I didn't search too far for a solution so took it as an opportunity to do some reverse engineering of a pcap.

I happen to have a couple different devices supported by the official Loupedeck software so will need to contribute over there as well.

@Terrorwolf01

Yes, that is the intent.

For the Stream Deck+, the layout model can describe the touch strip as a 'touchpanel' surface, with one or more logical 'Keypad' controls placed over it. In the fake-device plugin I’m using for testing, the SD+ strip is modeled as one continuous touch-panel surface with four logical touch zones.

For the Stream Deck Neo, the same model can represent the small non-LCD/touch-style controls separately from the main LCD key grid. The fake Neo layout uses a main display surface for the 2x4 LCD keys and a separate touch-panel/info-bar region for the lower controls.

So the model separates two concepts:

- 'surfaces': physical display or touch-panel regions, such as the SD+ strip
- 'controls': logical assignable positions, such as keys, encoders, or touch zones on top of a surface

That lets the UI show the physical shape while still routing drops/clicks to the correct logical position.

The exact hardware behavior still depends on the plugin that owns the device. This PR gives plugins a way to describe those regions to OpenDeck; it does not by itself add built-in hardware support for non-Elgato devices.

Terrorwolf01:
> The exact hardware behavior still depends on the plugin that owns the device. This PR gives plugins a way to describe those regions to OpenDeck; it does not by itself add built-in hardware support for non-Elgato devices.

Ofc. Never thought so. Was just checking. For the Touchpanel name: Wouldn't it make more sense to keep the same naming? In this case Touchstrip since thats the official Name in the Elgato Universe and in OpenDeck?

BrendanGrant:
Had to re-write a little history as the wrong email address got used for the commit.

@Terrorwolf01 

That's a fair point.

I used 'touchpanel' because I was trying to describe the generic shape of the capability rather than the Stream Deck+ specifically. The intent was to cover things like the SD+ touch strip, Neo-style touch areas, and non-Elgato devices that may have larger or differently shaped touch/display regions.

That said, consistency with existing OpenDeck / Elgato terminology matters. If 'touchstrip' is already the established term in OpenDeck, I'm fine renaming the layout surface kind to that, or using 'touchstrip' for the narrow horizontal SD+ case and reserving something more generic only if we later need it.

Being new here, my preference would be whatever fits the existing OpenDeck vocabulary best.

4ndv:
Is there a direct migration path from defining strips as set of buttons with separate ids?

I have a plugin for this device, with a strip on the side displayed as three distinct buttons right now:

image

Terrorwolf01:
> Is there a direct migration path from defining strips as set of buttons with separate ids?
> 
> I have a plugin for this device, with a strip on the side displayed as three distinct buttons right now:

Before thinking about a "Migration Path", this first needs to be finished in itself.
Other than that, PROBABLY the Profiles using the new Features would then in turn be remade since its probably not possible to just copy them over 1:1.

tommasobbianchi:
I've just closed #429, which tried to solve a narrower version of this with a repositioning flag. @nekename's objection there and on #427 is right, so let me offer the device instead as a test case — it exercises a corner I don't think the draft covers yet.

**VSD Stream Dock N1** (HOTSPOTEK 5548:1002, sold as VSDinside / TreasLin / ActionRing). It's a numpad-shaped device, so it stands in portrait:

```
+----------------------+
|  [64]   [64]   (64)  |   <- strip: 2 buttons + the knob's screen
|  [ 96 ] [ 96 ] [ 96 ]|
|  [ 96 ] [ 96 ] [ 96 ]|
|  [ 96 ] [ 96 ] [ 96 ]|   <- 15 main LCD keys, 3 x 5
|  [ 96 ] [ 96 ] [ 96 ]|
|  [ 96 ] [ 96 ] [ 96 ]|
+----------------------+
```

- 15 main keys, each its own **96x96** JPEG target, in 3 columns x 5 rows.
- Above them a strip of three **64x64** screens. The first two have a physical button behind them and report key codes. The third has **no button** — it is the screen the rotary encoder sits under.

So the encoder is not below the keypad and not on a shared touch strip: it is *inline with two keys*, above the block, and it owns a discrete square screen of its own.

Three things the legacy model can't say, and how I read the draft handling them:

1. **Encoder above the keypad, in line with keys.** `controls` with explicit coordinates covers this cleanly. This was the whole of #429 and the layout model subsumes it.
2. **Mixed key sizes on one device** (96x96 vs 64x64). Covered geometrically by per-control `width`/`height`.
3. **An encoder with its own dedicated display.** This is the gap. `DeviceLayoutControl::Encoder` takes `cx`/`cy`/`r` and an optional `surface`, but a surface is either `display` (grouping metadata) or `touchpanel` (drawn as a region). Neither says "this encoder has a 64x64 panel you can push an image to". On the SD+ the encoders share one strip, so the question doesn't arise; here it does.

Sketch of how I'd want to register it, assuming per-control pixel dimensions existed:

```json
"layoutVersion": 1,
"layout": {
  "canvas": { "width": 336, "height": 660, "unit": "px" },
  "surfaces": [
    { "id": "strip", "kind": "display", "x": 28, "y": 24, "width": 280, "height": 64 },
    { "id": "keys",  "kind": "display", "x": 12, "y": 108, "width": 312, "height": 528 }
  ],
  "controls": [
    { "controller": "Keypad", "position": 0, "shape": "rect", "x": 28,  "y": 24, "width": 64, "height": 64, "surface": "strip" },
    { "controller": "Keypad", "position": 1, "shape": "rect", "x": 136, "y": 24, "width": 64, "height": 64, "surface": "strip" },
    { "controller": "Encoder", "position": 0, "shape": "circle", "cx": 276, "cy": 56, "r": 32, "surface": "strip" },

    { "controller": "Keypad", "position": 2, "shape": "rect", "x": 12,  "y": 108, "width": 96, "height": 96, "surface": "keys" },
    { "controller": "Keypad", "position": 3, "shape": "rect", "x": 120, "y": 108, "width": 96, "height": 96, "surface": "keys" }
    // ... 13 more
  ]
}
```

Two questions that fall out of writing that:

- **Pixel dimensions belong on the control, not only the surface.** `pixelWidth`/`pixelHeight` currently live on the surface, but here a "surface" is a grouping of 15 physically separate panels, so there is no single pixel size for it. Would you consider optional `pixelWidth`/`pixelHeight` on a control, falling back to the surface's when absent? That also gives the renderer the right resolution to ask the plugin for per key.
- **Should an `Encoder` be allowed to declare a display?** Either an optional `pixelWidth`/`pixelHeight` on the encoder control, or letting it reference a `display` surface that is exactly its own panel. Without one of those, an encoder-with-a-screen has to be registered as a Keypad that happens to rotate, which is what I'd rather stop doing.

@4ndv's migration question is mine too, from the other side: my plugin registers that strip as a **3x6 grid whose row 0 is the strip**, which is exactly the "strip as a set of separate button ids" shape he described. Existing profiles are keyed by grid position, so whatever the migration story is, it needs to survive positions being renumbered when a device moves from the legacy renderer to an explicit layout — or explicitly keep `position` stable and let it stay an opaque index, which is what I'd prefer.

Happy to test the branch against real hardware — the plugin is at https://github.com/tommasobbianchi/opendeck-vsd-n1 and the device is on my desk. If the shape of the two additions above sounds acceptable I can also try implementing them on top of this PR.

PR #361, opened 2026-06-23
fix: reapply device brightness after system resume

sjourdois:
### Preflight checklist

---

## Summary

After the machine resumes from suspend, a connected Stream Deck's screen stays **black**
(brightness 0%) until the user manually moves the brightness slider.

**Root cause.** On resume, the device firmware resets its brightness to 0, but the OS does
*not* re-enumerate the device — OpenDeck's HID handle and software state remain valid (the
read loop in `elgato::init` merely keeps timing out, never erroring, so the device is never
torn down and re-initialised). Power monitoring (added in #325 via `psp`) already detects the
resume, but the `PowerState::Resume` handler only broadcasts `systemDidWakeUp` to plugins —
nothing re-pushes brightness to the devices. So the panel stays dark.

This follows up on the discussion in #357, where re-doing device wake recovery "using the
proper power monitoring afforded by psp" was suggested.

## Changes

- **`device_sleep.rs`**: add `reapply_brightness_after_resume()`, which re-pushes the
  configured brightness to every connected device that is not intentionally asleep
  (`is_device_sleeping` — i.e. idle-timeout or computer-lock sleep are respected and left off).
- **`power_events.rs`**: call it from the `PowerState::Resume` arm, alongside the existing
  `systemDidWakeUp` broadcast.

It re-uses `set_device_brightness`, so it also covers non-Elgato plugin devices. It writes
only to already-open handles — it never enumerates or opens a device — so it stays clear of
the macOS wake path being addressed in #354.

## Testing

- Linux (Wayland + systemd-logind), Stream Deck Plus.
- Before: real `systemctl suspend` / resume → screen stays black (reproduced on current
  `main`).
- After: real suspend / resume → screen returns to the configured brightness within ~1 s,
  without any user interaction. Confirmed in the log:
  `Reapplied brightness to 1 device(s) after system resume`.
- `cargo fmt` applied; no unrelated changes.

Not yet tested on Windows/macOS; the change is platform-agnostic (driven by `psp`) and avoids
the macOS enumerate/open path, so I do not expect platform-specific issues, but confirmation
on those platforms is welcome.

### DIFF
--- a/src-tauri/src/device_sleep.rs
+++ b/src-tauri/src/device_sleep.rs
@@ -16,6 +16,26 @@ pub fn is_device_sleeping(device: &str) -> bool {
 	SLEEPING_DEVICES.contains_key(device)
 }
 
+/// After the system resumes from sleep, device firmware has reset its brightness to 0, but
+/// OpenDeck's handles and software state are intact (devices are not re-enumerated). Re-push the
+/// configured brightness to every connected device that is not intentionally asleep (idle or
+/// computer-lock). Does not open/enumerate any device, so it is safe with respect to the macOS
+/// wake crash (which occurs in `hid_open`/`hid_enumerate`, not when writing to an open handle).
+pub async fn reapply_brightness_after_resume() -> Result<(), anyhow::Error> {
+	let brightness = crate::store::get_settings().value.brightness;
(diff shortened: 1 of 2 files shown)


nekename:
As it stands, this PR is in violation of section 1 of the General Guidelines of the LLM/AI policy, and will be closed as such.

PR #376, opened 2026-06-30
Handle image path expansion and validation inside OpenDeck

FrostyCoolSlug:
### Preflight checklist

- [X] I reasonably believe that the changes that I am proposing are of production quality, or I am otherwise opening this pull request as a draft.

---
This moves the path expansion and validation from the render lib to OpenDeck.

A bug was found where loading a custom icon of an action on startup loads the icon from `.config/opendeck/images/SD-.../x.svg`, which when passed in to the renderer was causing an 'Attempted to load image outside of base path' error.

This code moves the validation out of the renderer and into OpenDeck as it's better suited to make the determination over whether a path is acceptable or not.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -5394,7 +5394,7 @@ checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 [[package]]
 name = "streamdeck-strip-render"
 version = "0.1.0"
-source = "git+https://github.com/FrostyCoolSlug/streamdeck-strip-render?rev=5a18406c#5a18406c3f7560f4f4220756e7da459d76559946"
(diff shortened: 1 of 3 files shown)


sjourdois:
Hi,

The default encoder icon also fails to render. Its stored value is plugins/<uuid>/icons/x.svg (relative to the config dir), but get_encoder_image joins it with config_dir/plugins/\<uuid\> → …/plugins/\<uuid\>/plugins/\<uuid\>/… → not found.

Root cause looks like the render base being the plugin dir while stored image paths are relative to the config dir — using config_dir() as the base would resolve both the custom-icon (images/SD-…) and default-icon (plugins/<uuid>/…) cases

FrostyCoolSlug:
Can you please provide a plugin where this problem happens? The entire point of this PR is to solve this, and I can't reproduce a case where the default icon cannot be loaded.

sjourdois:
Oops... it seems my code was wrong, or maybe your PR really fixes it. I can't reproduce anymore.

FrostyCoolSlug:
Closing, superseded by #385

PR #377, opened 2026-07-01
fix: always answer getSettings

mja00:
### Preflight checklist

---
This makes `getSettings` always reply even if we're in a fresh context. Some plugins, in my case https://github.com/ciderapp/CiderDeck, unconditionally expect this to reply (and it appears the official StreamDeck program _does_) and will get stuck in a crash loop when it doesn't exist. 

Once the plugin + OpenDeck are fully initialized it's fine, just initial boot dies.

### DIFF
--- a/src-tauri/src/events/inbound/mod.rs
+++ b/src-tauri/src/events/inbound/mod.rs
@@ -80,7 +80,6 @@ pub async fn process_incoming_message(data: Result<Message, Error>, uuid: &str, 
 		if !(uuid.is_empty() && skip_auth) {
 			if let Some(context) = match &decoded {
 				InboundEventType::SetSettings(event) => Some(&event.context),
-				InboundEventType::GetSettings(event) => Some(&event.context),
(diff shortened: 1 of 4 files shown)


nekename:
This is really weird. The only way that the event wouldn't be sent is if it was requesting settings for an action that didn't exist or if it's requesting settings for an action from a different plugin. Don't you think this is an issue that should be raised with the plugin developer first?

If they refuse to fix it, then I can consider merging this. But I think you should at least ask them first.

mja00:
> This is really weird. The only way that the event wouldn't be sent is if it was requesting settings for an action that didn't exist or if it's requesting settings for an action from a different plugin. Don't you think this is an issue that should be raised with the plugin developer first?
> 
> If they refuse to fix it, then I can consider merging this. But I think you should at least ask them first.

Yeah I'm drafting up a PR for there as well. It's odd that the windows StreamDeck software just will blindly do this, since it _does_ offer potential wonkiness.

nekename:
Let me know if this needs to be reopened

PR #381, opened 2026-07-02
fix: ignore --hide relaunches in single-instance callback to respect background startup

### LINKED ISSUE #380: App opens in windowed mode after login
## Bug description

OpenDeck always opens with a visible window after logging into the desktop session, even when "Run in background" is enabled in Settings.

## Root cause

The "Run in background" setting (`settings.background`) only ever controlled what happens when the main window's close button is clicked (minimise to tray vs. quit) — it was never consulted at startup. Startup visibility is instead governed purely by the `--hide` CLI argument that the autostart integration passes (`tauri_plugin_autostart::init(..., Some(vec!["--hide"]))`).

If OpenDeck is already running when the autostart entry fires a second launch with `--hide` (e.g. an instance from a previous session still running, or DE-specific autostart/session-restore timing — this seems more likely on KDE Plasma sessions, similar to the class of issue fixed in 0ca1c3f), the `tauri_plugin_single_instance` callback in the already-running process receives the relaunch args. None of its recognised argument branches (`--reload-plugin`, `--sleep-device`, etc.) matched `--hide`, so it fell through to the final `else` branch, which unconditionally called `show_window(app)` — forcing the window open regardless of the "Run in background" setting.

## Suggested fix

The single-instance callback should ignore `--hide`-only relaunches instead of forcing the window visible:

```rust
} else if !args.iter().any(|x| x.to_lowercase().trim() == "--hide") {
    let _ = show_window(app);
}
```

The "Run in background" tooltip is also misleadingly worded, since it only affects close-button behaviour, not startup visibility — worth clarifying alongside the fix.

I have a working fix for this (tested locally, built and verified with `--hide` both on first launch and on relaunch while an instance is already running) and am happy to open a PR if useful.

l0bster:
Fixes #380

## Summary

- Autostart launches OpenDeck with `--hide`. If an instance is already running (e.g. leftover from a previous session, or DE-specific autostart/session-restore timing — more likely on KDE Plasma), the `tauri_plugin_single_instance` callback receives the relaunch args in the already-running process.
- None of the callback's recognised argument branches matched `--hide`, so it fell through to the final `else`, which unconditionally called `show_window(app)` — forcing the window visible regardless of the "Run in background" setting.
- Also clarified the "Run in background" tooltip (EN + DE), which only ever governs close-button behaviour (minimise to tray vs. quit), not startup visibility — the previous wording implied it controlled whether the app starts hidden.

## Test plan

- [x] `cargo clippy` / `cargo fmt` clean
- [x] `deno lint` / `deno task check` / `deno task format` clean
- [x] Built release binary locally, ran with `--hide` as first instance — window stays hidden (verified via temporary debug logging, removed before commit)
- [x] Ran a second `--hide` launch while the first instance was still running — confirmed the single-instance callback now skips `show_window` instead of forcing the window open

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -346,7 +346,7 @@ If you have already donated, thank you so much for your support!"#,
 								tauri::async_runtime::block_on(events::inbound::process_incoming_message(Ok(tokio_tungstenite::tungstenite::Message::Text(message.into())), "", true));
 							});
 						}
-					} else {
+					} else if !args.iter().any(|x| x.to_lowercase().trim() == "--hide") {
 						let _ = show_window(app);
 					}
 				})
--- a/translations/de.json
+++ b/translations/de.json
@@ -114,7 +114,7 @@
 	"settings.autolaunch.tooltip.1": "Wenn diese Option aktiviert ist, wird {{PRODUCT_NAME}} automatisch beim Anmelden gestartet.",
 	"settings.autolaunch.tooltip.2": "Wenn du {{PRODUCT_NAME}} mit Flatpak installiert hast, funktioniert diese Option möglicherweise nicht wie vorgesehen",
 	"settings.background": "Im Hintergrund weiterlaufen lassen:",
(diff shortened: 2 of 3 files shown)


Terrorwolf01:
This will be closed to not following the Checklist. (And probably not following the policies of OpenDeck for PRs mentioned in the Checklist)

Issue #152, opened 2025-09-26
Custom icons being overwritten by plugin events


mkstephens:
I am a new OpenDeck user with a new Stream Deck MK.2. I am using the Multi OBS Controller Plugin. I am able to customize the icon for a button, however if I disconnect the Stream Deck and then reconnect it, it reverts back to the default icon. I don't need to use the Stream Deck all the time, so I anticipate that this will be a pretty big issue for me as being able to customize the icons is necessary to keep all the actions straight.
I am using Manjaro Linux on a Framework 13 laptop.
Thanks in advance for your help.

nekename:
This had already been reported on the Discord, I think. I don't have unlimited resources, but I am aware of it (it is not specifically about this plugin) and someday it will be fixed.

MiguVT:
Tried with OpenDeck run command option and dont had this issue, maybe releated to other thing that makes or the plugin have this problem or plugins not releated to opendeck and directly with streamdeck.

Using arch linux (CachyOS) with KDE Plasama (Desktop not releated if im not wrong?)

nekename:
@MiguVT I'm not sure I understand. You *don't* have this issue? Then why are you commenting here?

MiguVT:
> [@MiguVT](https://github.com/MiguVT) I'm not sure I understand. You _don't_ have this issue? Then why are you commenting here?

Trying to give some information, I would love to try help find the root issue <3

You know maybe its releated to a weird issue that only happens on x model of streamdeck or with extensions that x you know how that things works 😅

nekename:
> You know maybe its releated to a weird issue that only happens on x model of streamdeck or with extensions that x you know how that things works 😅

Yes, but normally only people that actually have the issue in question can provide any useful information, lol...

MiguVT:
> > You know maybe its releated to a weird issue that only happens on x model of streamdeck or with extensions that x you know how that things works 😅
> 
> Yes, but normally only people that actually have the issue in question can provide any useful information, lol...

Yes but just trying to give information of which environment dont have this issue I can also try to replicate the issue if someone gives more information (and also contribute something if needed)

nekename:
I see. If you want to be someone who helps with troubleshooting lots of different issues with OpenDeck, please do join the Discord, observe for a month or so, then help people out with their issues, as many more issues are reported there than here on GitHub.

pshempel:
[...]

PR #387, opened 2026-07-10
fix: preserve plugin-set key images across profile re-renders

### LINKED ISSUE #386: Plugin-set key image is reverted to the profile's loaded value when another action is added or moved
### Preflight checklist

**If you remove this checklist, this issue will be closed without explanation.**

- [x] I understand that if this issue is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38, and that it should be raised in the repository of the plugin that provides support for the hardware.
- [x] I will provide the OpenDeck log file and the log files of plugins involved, where applicable.
- [x] I will provide my operating system information and method of installation, as well as any steps I have taken to troubleshoot the issue.
- [x] I certify that I am using the latest available version of OpenDeck, and that I will provide enough information to reproduce the issue.

---

### Description

A plugin updates a key's image with `setImage` during a session, but the update is
silently discarded on the next profile grid re-render (e.g. after adding or moving any
action). The key then reverts to the image the profile was loaded with. No device
reconnect is involved.

### Steps to reproduce

1. Add a plugin action that paints its key image via `setImage`, set an image on it,
   and let OpenDeck save the profile.
2. Restart OpenDeck. The key shows the saved image (loaded from the profile).
3. Have the plugin set a **different** image via `setImage` this session — the key
   updates on screen. 
4. Add or move any other action anywhere on the profile.
5. The key reverts to the image from step 2, losing the update from step 3. 

### Expected behavior

The image set in step 3 persists across the re-render.

### Actual behavior

The key reverts to the value the profile was loaded with on the next grid re-render.

### Relationship to #152

This shares the symptom of #152 (a key showing the wrong image) but appears to be a
different cause. #152's reported scenario is a device disconnect/reconnect, where the
in-memory profile is dropped and reloaded from disk (a backend-side path). The case
here is a purely frontend re-render that happens with no reconnect.

### Environment

- OpenDeck version: v2.13.1 (built from source, `main`)
- Method of installation: built from source (`deno task tauri dev`)
- Operating system: Windows 11 Pro 25H2
- Logs: not applicable — deterministic UI-state bug, reproducible from the steps above

atomskz:
### Preflight checklist

---

Fixes #386.

> Related to #152 — the same symptom (a key showing the wrong image) but a distinct
> cause. #152 is a device reconnect (a backend reload from disk); this fixes a purely
> frontend re-render and does not address the reconnect case.

### Problem

When a plugin updates a key's image via `setImage`, the update is discarded on the
next profile grid re-render (e.g. after adding or moving any action), and the key
reverts to the image the profile was loaded with. No reconnect is involved.

### Cause

`slot` is a Key's local working copy; `inslot` is the deliberate one-way downward
channel from `profile.keys[]` (a plugin's per-session `setImage` writes only `slot`,
never `profile`). On `profile = profile`, Svelte's `safe_not_equal` treats an
identical object reference as changed, so the reactive `$: update(inslot)` re-runs for
every key even when its `profile.keys[i]` didn't change; `update()` then
unconditionally does `slot = inslot`, resetting the key to the loaded instance and
discarding the plugin's image.

### Fix

Make `update()` idempotent: remember the last bound reference and skip the reset when
`inslot` hasn't actually changed. The one-way `inslot`/`slot` split stays intact —
nothing propagates upward into `profile` — while a genuine change (a fresh instance
from the backend, or the `update_state` listener) still applies.

```diff
 	// One-way binding for slot data.
 	export let inslot: ActionInstance | null;
 	let slot: ActionInstance | null;
+	let lastInslot: ActionInstance | null | undefined;
 	const update = (inslot: ActionInstance | null) => {
+		if (inslot === lastInslot) return;
 		if (inslot && context && inslot.context.split(".")[0] != context.device) return;
+		lastInslot = inslot;
 		slot = inslot;
 	};
 	$: update(inslot);
```

The latch sits after the device filter, so a filtered entry is re-evaluated on the
next pass exactly as before, and `clear()`'s upward `null` still round-trips.

### Testing

- With a plugin that updates a key image via `setImage`: set an image, restarted so
  the profile loads it, set a different image this session, then added/moved another
  action — the update now persists (previously the key reverted to the loaded image).
- Keys whose image matches the loaded profile continue to render correctly.
- `deno task check` and prettier are clean.

### DIFF
--- a/src/components/Key.svelte
+++ b/src/components/Key.svelte
@@ -27,8 +27,11 @@
 	// One-way binding for slot data.
 	export let inslot: ActionInstance | null;
 	let slot: ActionInstance | null;
+	let lastInslot: ActionInstance | null | undefined;
 	const update = (inslot: ActionInstance | null) => {
+		if (inslot === lastInslot) return;
 		if (inslot && context && inslot.context.split(".")[0] != context.device) return;
+		lastInslot = inslot;
 		slot = inslot;
 	};
 	$: update(inslot);


nekename:
This is a fix already discovered by @justmangoou, but I'm wary of accepting it because there must have been some reason for separation of `inslot` and `slot` - `inslot` specifically exists so that changes in Key don't propagate upwards. 

Anyway, if it turns out that `inslot` is actually no longer necessary, it should be removed entirely so that the changes made in InstanceEditor are also propagated upwards into the `profile` so that they are not reset on re-render.

And finally, the ultimate fix is of course to actually discover why other keys are being re-rendered when they're not supposed to. It causes unnecessary churn in the entire frontend and of course causes device updates. If we prevent the unnecessary re-renders of other keys when one key is added/deleted (which seems to be the main trigger of this issue), this won't be necessary at all.

Lastly, I'd rather read issue comments from a human than an LLM, if that's possible.

atomskz:
Hi @nekename, thanks for taking the time to review this!

**On your suspicion** — you're right, and I checked the history to be sure. The `inslot`/`slot` split landed in b2c5014 ("Implement user state modification") together with the "One-way binding for slot data" comment, so the one-way design is clearly deliberate: `slot` is the local working copy, `inslot` is the downward channel from `profile.keys[]`, and the only intended upward write is the explicit `inslot = slot` in `clear()`. A plugin's per-session `setImage` is runtime state, not configuration, so it has no business leaking into `profile`. That invariant is worth keeping — which is why I now agree the PR as submitted is wrong. (For what it's worth, the revert has been possible ever since that split landed: `update_state` has only ever written to `slot`, so any grid re-render could clobber it.)

**Why the write-back is incorrect.** `inslot = payload.contents` does the one thing the split exists to prevent: it pushes transient plugin state upward into `profile.keys[]`. It's also inconsistent — InstanceEditor still writes only `slot`, as you pointed out — and it doesn't reduce any churn: every other key still re-renders and re-pushes its image to the device on each add/move; the write-back just makes the re-pushed value happen to be right. It patches the symptom while breaking the design.

**The replacement.** I've reworked the change to leave the one-way design untouched and instead make `update()` idempotent — skip the reset when the bound reference didn't change:

```diff
+ let lastInslot: ActionInstance | null | undefined;
const update = (inslot: ActionInstance | null) => {
+	if (inslot === lastInslot) return;
	if (inslot && context && inslot.context.split(".")[0] != context.device) return;
+	lastInslot = inslot;
	slot = inslot;
};
```

The mechanism: on `profile = profile`, Svelte's `safe_not_equal` treats an identical object reference as changed, so `$: update(inslot)` re-runs for every key even when its `profile.keys[i]` didn't change, and the unconditional `slot = inslot` is what clobbers the plugin-set image. With the guard, an untouched key early-returns: the image survives and nothing propagates upward. The reference check is safe because a real change always arrives either as a fresh object from the backend (`create_instance` / `move_instance` / paste responses, profile refetches — I went through every write to `profile.keys[]` / `sliders[]` / `infobars[]`) or directly through the `update_state` listener. `clear()`'s upward null still round-trips correctly, and the latch sits after the device filter, so a filtered entry gets re-evaluated on the next pass exactly like upstream. `deno task check` and prettier are clean (the two `Timeout` errors in Key.svelte are pre-existing).

Compared to the write-back: same bug fixed, but nothing flows upward, the InstanceEditor inconsistency isn't made worse, and the runtime-vs-configured divergence stays where the design puts it — in `slot`.

**On the re-renders themselves** — I dug into why other keys re-render at all, and there are two layers. The `slot` clobbering is what the guard fixes. But the frontend/device churn you describe has a second, independent cause: DeviceView passes `context={{ device: device.id, ... }}` as an inline literal, so every `profile = profile` hands every Key a brand-new `context` object, and the render block depends on `context` — so all keys still re-render and re-push to the device even with the guard in place. Fixing that would mean hoisting/memoizing the per-slot context objects in DeviceView (their contents only change on device/profile switch). Happy to take that on as a follow-up — with both pieces in place, an add/move would touch only the affected key, which I believe is the "ultimate fix" you described.

If you'd rather go the remove-`inslot` route instead, or you and @justmangoou already have something in flight, no hard feelings — I'll close. Otherwise I'll update the PR to the guard version.

> Lastly, I'd rather read issue comments from a human than an LLM, if that's possible.

Understood, and fair enough. I use models to format and translate my comments into English, since it isn't my native language.

nekename:
Excellent work, thank you. That sounds like an acceptable fix that doesn't require changes in DeviceView. I'm away for a few more days this month but I'd be happy to test and merge this when I get back.

As for LLM usage for translation, that's perfectly OK but you should just mention that you're using it for translation so that it doesn't come across the wrong way.

nekename:
Looks good and works well; thank you @atomskz!

PR #393, opened 2026-07-17
feat: add delay in between multiaction children

KeryaneK:
### Preflight checklist

---

This PR introduces delays in multiaction. They're set to 0 by default to not break current user settings.
Also there are in-built delays - 100ms on keydown and 100ms on keyup. I haven't touched them as I'm not sure about their purpose. Delay settings are saved as an array in multiaction settings.


KeryaneK:
I've updated it to use parent's action settings to not interfere with plugins.

KeryaneK:
Basic manual tests done:

- settings do not override themselfs on update and reversion.
- delay action shows only when it's needed ( >2 actions, not after the last, not before the first)
- gaps are proper, even with delay actions in between them

Changing it to "Open" then although any improvement ideas are still welcome.

KeryaneK:
Force-push was about commit signing - no changes.

nekename:
@KeryaneK Thanks for the PR. I've just looked it over and while your changes themselves look good, we have a pretty large problem. It looks like the locks (mutable ones even!) are being held across the code for executing the children with delays, which, although awful, seems to have worked acceptably until now with the short delays, but this is going to be made much worse with longer delays.

Could you refactor this so that the locks are dropped before executing the children?

nekename:
Note: for keeping the automatic state flipping, you'll probably need to move it out of the loop and then get the locks again afterwards

KeryaneK:
Yup, my bad. That was indeed an issue.

Hope the current version does it properly in your opinion.
Thanks. :)

nekename:
Thank you, will look tomorrow (I think)

nekename:
@KeryaneK could you let me know if it's still good after my changes? I'm also wondering if we should include the 100ms delays in this field and allow the user to customise them instead of that being blocked.

@Terrorwolf01 @psoloviov @snowfruit @fcrespo82 could you provide translations for "Delay:" in your languages in the comments if possible please?

KeryaneK:
It works fine.

I havent touched delays that were hard-coded because I had no time to go though and check if they aren't needed somewhere. How would you see it? Just two additional fields with delay after down, delay after up or maybe a main delay (as it is now) splitted between them? So delay 500ms == 250ms on key down and 250ms on key up?

nekename:
I think it makes the most sense to keep the delay within each action hardcoded at 100ms, since that avoids complex UI and no one needs to adjust that value, then allow customising the delays between separate actions (i.e. completely, not with an extra hardcoded 100ms, but we can keep the default at 100ms).

snowfruit:
> @KeryaneK could you let me know if it's still good after my changes? I'm also wondering if we should include the 100ms delays in this field and allow the user to customise them instead of that being blocked.
> 
> @Terrorwolf01 @psoloviov @snowfruit @fcrespo82 could you provide translations for "Delay:" in your languages in the comments if possible please?

"Fördröjning:"

Terrorwolf01:
> @KeryaneK could you let me know if it's still good after my changes? I'm also wondering if we should include the 100ms delays in this field and allow the user to customise them instead of that being blocked.
> 
> could you provide translations for "Delay:" in your languages in the comments if possible please?

German Translation is ok.

nekename:
ok @KeryaneK I added the swedish, I fixed the broken margins in the toggle action UI, I added an ARIA label so that screen reader users know what the delay fields are, and then I allowed the built-in 100ms delay to be the default but to be customisable

let me know if it's good

nekename:
I'm also working on fixing the remove logic so that it preserves the delay for each remaining action as the one that was before it, but I'll have to do that later today

KeryaneK:
@nekename Is it on purpose that the user cannot set "0" as a delay?

Additionally, current fallback to 100 on NaN (empty field) makes it impossible to change the 1st digit using backspace. (so 100ms -> 10ms -> 1ms -> reverts to 100 again).

nekename:
Not being able to use 0 was a mistake, my bad there - hopefully fixed now.

~~As for it resetting to 100 when you clear the field, I think in your code it also reset to `0` when you cleared the field - do you have a suggestion for fixing that?~~

Edit: nvm, fixed that as well. Thanks

KeryaneK:
It was an issue previously - although not that drastic as it looked as it's doing that on purpos. When going back to 100ms it looked like a bug. :D

I already had it fixed (in the same way as you did lol), just without the push as I was waiting for your response. Cool that you did it yourself.
Feel free to let me know if anything more is in need of adjustments. ;)

nekename:
I've been spending forever trying to make it always remove the delay before a child when you remove it, and banging my head against the wall because I didn't remember that `context.index` isn't sequential after a child is removed :sob:

Spent way too long but I should push it soon

nekename:
Alright, 0187cc4c8b2279593a3a7aaf1dfd6f578cc7e58c, let me know what you think, this should be good to merge now!

KeryaneK:
It works nearly perfect. One thing I noticed is a bug where it assigns the delay of deleted action instead of the one which stayed:

Action 1
Delay 500ms
Action 2
Delay 1000ms
Action 3

---> Delete Action 2

Action 1
Delay 1000ms
Action 2

Now this commit fix it.

nekename:
Force pushed to revert that change bc I meant it as is. 

> because i was treating it as an action's delay is how long it waits before that action
like you're delaying that action by that amount
that's how the aria labels are set up as well

Thanks @KeryaneK - one more off the "bucket list"!

translations/en.json
```diff
@@ -41,6 +41,8 @@
 	"no_devices_detected.title": "No devices detected",
 	"parent_action_view.child": "child",
 	"parent_action_view.children": "children",
+	"parent_action_view.delay.aria": "Delay for {{name}}",
```
fcrespo82, line 44: “Atraso para {{name}}”

translations/en.json
```diff
@@ -41,6 +41,8 @@
...
+	"parent_action_view.delay.label": "Delay:",
```
fcrespo82, line 45: “Atraso:”

nekename:
Shoot, I completely forgot about the translations! I'll add those and the Ukrainian ones tomorrow (will wait for @psoloviov and I'll use AI if they haven't updated by then)

nekename:
21e714467f3b1d2b483adbbb520e7c5c85cd1ffd

psoloviov:
> [21e7144](https://github.com/nekename/OpenDeck/commit/21e714467f3b1d2b483adbbb520e7c5c85cd1ffd)

I reviewed this, everything fine.
Currently I have heavy load, so I didn't do this by myself

PR #405, opened 2026-07-31
feat: add Korean translation

jaehong5:
## Summary

Adds `translations/ko.json` with all 152 keys translated into Korean.

The language selector in `SettingsView.svelte` already offered a `ko` option, but no matching translation file existed. Since `i18n.ts` falls back to `FALLBACK_LOCALE` when a locale is missing, selecting Korean silently displayed English with no indication that anything was wrong.

## Label fix

The `ko` option was labelled `韓国語`. That is the *Japanese* word for the Korean language — most likely copied from the `ja` option directly above it. Every other option uses the language's own endonym (`Deutsch`, `Українська`, `中文`), so this is now `한국어`.

```diff
-<option value="ko">韓国語</option>
+<option value="ko">한국어</option>
```

## Translation notes

Two keys needed the surrounding markup to translate correctly, because the key names and the widgets they label are swapped:

- `instance_editor.stroke` labels an `<input type="color">` bound to `stroke_colour` → translated as "외곽선 색" (outline colour)
- `instance_editor.outline` labels an `<input type="number">` bound to `stroke_size` → translated as "두께" (thickness)

Translating these from the key names alone would have produced labels that do not match their controls.

Other choices worth noting:

- `parent_action_view.child` / `.children` are concatenated as `"{Multi Action} {child} {index}"`, so both are rendered as "하위 항목". Korean has no grammatical plural, so the two keys share a translation.
- `settings.footer.1`–`.5` are sentence fragments joined around two links. Korean word order differs from English, so the fragments were redistributed to read naturally while keeping the same link positions.
- "Action" is rendered as "액션" rather than a native-word equivalent, matching how the term is used across the Stream Deck plugin ecosystem.

## Verification

- 152/152 keys present, same order as `en.json`, no extras or omissions
- All `{{variable}}` placeholders preserved and matched against the source
- `\n` counts match `en.json`
- Tab indentation and LF line endings, consistent with the existing translation files

## Remaining untranslated locales

For context, the selector currently offers 10 languages but only 5 have translation files. After this PR, `es`, `fr`, `zh_CN` and `ja` still fall back to English the same way Korean did.


Terrorwolf01:
> The language selector in SettingsView.svelte already offered a ko option, but no matching translation file existed. Since i18n.ts falls back to FALLBACK_LOCALE when a locale is missing, selecting Korean silently displayed English with no indication that anything was wrong.

As stated in the Tooltip, this Dropdown is now used not only for Plugin Translation as was its intended use, but also for the Interface Translation. It also states, that not all Languages are available for the Interface. I will have a look if we Maybe add a Banner stating that the selected Language only has some Supported Plugins that are translated, and the Language is currently missing Support in openDeck.


> settings.footer.1–.5 are sentence fragments joined around two links. Korean word order differs from English, so the fragments were redistributed to read naturally while keeping the same link positions.

Thanks for the Info, I completely overlooked this when I decided to use this Approach.


> \n counts match en.json

The \n is only for next line, so it can differ if needed in your Language.


> For context, the selector currently offers 10 languages but only 5 have translation files. After this PR, es, fr, zh_CN and ja still fall back to English the same way Korean did.

Same as the First point, but since we don't use Translation engines but Rely on User Translated Strings, they are still open.

nekename:
> The \n is only for next line

They just mean that they've checked that there's the same number of translation keys.

I'm sorry about the Korean label using Japanese text. I'm not sure how that happened or how I could have overlooked it, considering I can read a little bit of Korean myself (I'm a bit of a linguist).

jaehong5:
Hi, I wanted to check whether there is anything I should change before this PR can be reviewed or merged. Thanks!

Terrorwolf01:
> Hi, I wanted to check whether there is anything I should change before this PR can be reviewed or merged. Thanks!

AFAIK Nekename is currently taking a bit of time off from OpenDeck but should be back on soon. (I believe he said somewhere during the summer he will take a break.)

PR #409, opened 2026-08-02
Fix/plugin profile switch clean

Matthias3199:
## Summary

This PR restores compatibility with the official Stream Deck SDK `switchToProfile` command.

Although OpenDeck already had internal support for profile switching, plugins using the official `@elgato/streamdeck` SDK could not successfully switch profiles.

## Changes

- Accept the official `switchToProfile` event name using a serde alias.
- Parse the official SDK payload format (`payload.profile` and `payload.page`).
- Allow regular plugins to call `switchToProfile`.
- Update the selected profile immediately in the backend.
- Notify the frontend after the profile has been changed.

## Why

The official Stream Deck SDK exposes the following API:

```ts
streamDeck.profiles.switchToProfile(deviceId, profile, page);
```

Before this change, the command could not be used correctly because:

- OpenDeck only accepted a different event name.
- The payload format was incompatible with the official SDK.
- The command was restricted to internal plugins.
- Receiving the event did not actually update the active profile.

As a result, plugins built against the official SDK were unable to switch profiles.

This PR makes the official SDK command work as expected while preserving the existing frontend behavior.

## Testing

Tested with a custom plugin using the official `@elgato/streamdeck` SDK.

Verified that:

- `switchToProfile()` is received correctly.
- the requested profile becomes the active profile.
- the frontend updates accordingly.
- profile switching works repeatedly in both directions.
- existing functionality continues to work normally.

## Notes

This PR is independent from my touchscreen swipe event PR.

It only improves compatibility with the official Stream Deck SDK.

### Preflight checklist

- [] I understand that if this pull request is about support for non-Elgato or non-Tacto hardware, it will be closed without explanation, as per issue #38.
- [] I certify the compliance of this pull request with the [Jellyfin LLM/"AI" Development Policy](https://web.archive.org/web/20260414131310/https://jellyfin.org/docs/general/contributing/llm-policies/) adopted by the OpenDeck project.
- [] I thoroughly understand the changes that I am proposing in this pull request, and I will be able to respond to questions and review feedback.
- [] I have thoroughly reviewed the diff of my changes and ensured that I have neither introduced any unrelated additions, nor differences in unmodified code.
- [] I have ensured that I have run the appropriate formatter on my changes and that my code produces no linter violations.
- [] I will keep "Allow edits from maintainers" enabled for this pull request.

---

### DIFF
--- a/src-tauri/src/events/inbound/misc.rs
+++ b/src-tauri/src/events/inbound/misc.rs
@@ -43,14 +43,38 @@ pub async fn show_ok(event: ContextEvent) -> Result<(), anyhow::Error> {
 }
 
 #[derive(Clone, Serialize, Deserialize)]
+pub struct SwitchProfilePayload {
+	profile: Option<String>,
+	page: Option<u32>,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
 pub struct SwitchProfileEvent {
 	device: String,
-	profile: String,
+	payload: SwitchProfilePayload,
 }
 
 pub async fn switch_profile(event: SwitchProfileEvent) -> Result<(), anyhow::Error> {
+	let Some(profile) = event.payload.profile.clone() else {
+		anyhow::bail!("switchToProfile without a profile is not supported yet");
+	};
+
+	eprintln!("[profile-switch] reçu device={} profile={} page={:?}", event.device, profile, event.payload.page);
+
+	crate::events::frontend::profiles::set_selected_profile(event.device.clone(), profile.clone()).await?;
+
+	eprintln!("[profile-switch] succès device={} profile={}", event.device, profile);
+
 	let app_handle = crate::APP_HANDLE.get().unwrap();
-	app_handle.get_webview_window("main").unwrap().emit("switch_profile", event)?;
+
+	app_handle.get_webview_window("main").unwrap().emit(
+		"switch_profile",
+		serde_json::json!({
+				"device": event.device,
+				"profile": profile
+		}),
+	)?;
+
 	Ok(())
 }
 
--- a/src-tauri/src/events/inbound/misc.rs
+++ b/src-tauri/src/events/inbound/misc.rs
@@ -59,11 +59,9 @@ pub async fn switch_profile(event: SwitchProfileEvent) -> Result<(), anyhow::Err
 		anyhow::bail!("switchToProfile without a profile is not supported yet");
 	};
 
-	eprintln!("[profile-switch] reçu device={} profile={} page={:?}", event.device, profile, event.payload.page);
+	log::debug!("Switching profile for device {} to {} on page {:?}", event.device, profile, event.payload.page);
 
 	crate::events::frontend::profiles::set_selected_profile(event.device.clone(), profile.clone()).await?;
-
-	eprintln!("[profile-switch] succès device={} profile={}", event.device, profile);
 
 	let app_handle = crate::APP_HANDLE.get().unwrap();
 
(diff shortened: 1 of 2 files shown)


Terrorwolf01:
Not completly sure, but I think @nekename siad he did this on Purpose, because the Switch Profile event is mostly (maybe only) thought for Plugin supplied profiles which are 1) not Supported by OepnDeck and 2) can't work with OpenDeck due to Profiles not beeing compatible.

Matthias3199:
Thanks for the clarification. That makes sense.

I discovered this while building a plugin that uses the official @elgato/streamdeck SDK. I wanted the plugin to navigate between existing OpenDeck user profiles, so I initially reused switchToProfile.

I understand now that the official event is intended for plugin-supplied profiles, which OpenDeck does not currently support, and that mapping it directly to OpenDeck user profiles changes its original semantics.

Would you prefer one of these approaches?

Keep switchToProfile restricted until plugin-supplied profiles are supported.
Introduce a separate OpenDeck-specific event for switching OpenDeck user profiles.
Expose profile switching through another OpenDeck extension/API intended specifically for user profiles.

I am happy to adjust or close this PR depending on the direction you prefer.

Terrorwolf01:
> Thanks for the clarification. That makes sense.
> 
> I discovered this while building a plugin that uses the official @elgato/streamdeck SDK. I wanted the plugin to navigate between existing OpenDeck user profiles, so I initially reused switchToProfile.
> 
> I understand now that the official event is intended for plugin-supplied profiles, which OpenDeck does not currently support, and that mapping it directly to OpenDeck user profiles changes its original semantics.
> 
> Would you prefer one of these approaches?
> 
> Keep switchToProfile restricted until plugin-supplied profiles are supported. Introduce a separate OpenDeck-specific event for switching OpenDeck user profiles. Expose profile switching through another OpenDeck extension/API intended specifically for user profiles.
> 
> I am happy to adjust or close this PR depending on the direction you prefer.

For now I would say, leave this as is, a Profile Rewrite is going on right now as far as I am Informed (according to some things discussed in the discord). Maybe afterwards its easier to make some compatiblity to Plugin supplied Profiles. Then we can think about this.

But we can wait what @nekename says about this before closing the PR if you want.

Matthias3199:
> > Thanks for the clarification. That makes sense.
> > I discovered this while building a plugin that uses the official @elgato/streamdeck SDK. I wanted the plugin to navigate between existing OpenDeck user profiles, so I initially reused switchToProfile.
> > I understand now that the official event is intended for plugin-supplied profiles, which OpenDeck does not currently support, and that mapping it directly to OpenDeck user profiles changes its original semantics.
> > Would you prefer one of these approaches?
> > Keep switchToProfile restricted until plugin-supplied profiles are supported. Introduce a separate OpenDeck-specific event for switching OpenDeck user profiles. Expose profile switching through another OpenDeck extension/API intended specifically for user profiles.
> > I am happy to adjust or close this PR depending on the direction you prefer.
> 
> For now I would say, leave this as is, a Profile Rewrite is going on right now as far as I am Informed (according to some things discussed in the discord). Maybe afterwards its easier to make some compatiblity to Plugin supplied Profiles. Then we can think about this.
> 
> But we can wait what @nekename says about this before closing the PR if you want.

Thanks, 

I agree that it is better to wait for the ongoing profile rewrite before deciding how profile switching should be exposed to plugins.

I am happy to leave the PR open for now and wait for @nekename’s opinion. If the new profile architecture changes the preferred approach, I can update or close the PR accordingly.

nekename:
This event is indeed only for plugin-supplied profiles. OpenDeck has its own internal event that is currently limited to use by the Starter Pack only in order to prevent third-party plugins causing unexpected behaviour. I don't believe that switching between the user's own profiles without a clear indicator of what caused the switch will ever be a good UX.

Issue #62, opened 2025-03-06
Window not large enough to view debug info when Settings popup is open


MarcinOrlowski:
The `Settings` window should perhaps be resized as with default size it does not really fits all the texts properly:

![Image](https://github.com/user-attachments/assets/91d030d8-440f-4112-b5b7-5e3b504bae59)

I run `opendeck_2.4.2_amd64.AppImage` on Ubuntu 24.04.2

ninjadev64:
Well, the window size depends on the size of your device.

And also that's just debug info anyway and is collapsed by default, if you want to see it you can just enlarge the window.

PR #412, opened 2026-08-05
fix: stop discarding key image quality on the way to the device

### LINKED ISSUE #411: Key images lose their antialiasing on the way to the device
Key images arrive on the panel with their antialiasing destroyed — stair-stepped diagonals, strokes whose width alternates along their length, and small text that turns to mush. I traced this while writing an ALSA plugin that draws its key faces as SVG, on OpenDeck 2.14.0 with a Stream Deck +.

The path a key image takes:

1. `renderImage` draws it into a 144×144 canvas and hands it over as `canvas.toDataURL("image/jpeg")` — [`src/lib/rendererHelper.ts`](https://github.com/nekename/OpenDeck/blob/main/src/lib/rendererHelper.ts). Lossy pass 1.
2. `update_image` decodes it and calls `set_button_image` — [`src-tauri/src/elgato.rs`](https://github.com/nekename/OpenDeck/blob/main/src-tauri/src/elgato.rs).
3. `elgato-streamdeck` runs `resize_exact(120, 120, FilterType::Nearest)`, then re-encodes at JPEG quality 90. Lossy pass 2.

Step 3 is the root cause and I've filed it separately: OpenActionAPI/rust-elgato-streamdeck#61. 144 → 120 is a 6:5 ratio, so nearest-neighbour drops every sixth row and column and point-samples the antialiased edges instead of averaging them. Re-running the identical pipeline with `Lanczos3` in place of `Nearest` and changing nothing else produces a clean image, so it isn't the JPEG passes or the artwork.

Two things OpenDeck could do without waiting on a crate release.

**Neutralise the nearest-neighbour resize locally.** If the image already matches the device's key size, `resize_exact` has nothing left to do:

```rust
// src-tauri/src/elgato.rs, in update_image
} else {
    let img = image::load_from_memory(&bytes)?;
    let (w, h) = kind.key_image_format().size;
    let img = if img.width() != w as u32 || img.height() != h as u32 {
        img.resize_exact(w as u32, h as u32, image::imageops::FilterType::Lanczos3)
    } else {
        img
    };
    device.set_button_image(context.position, img).await?;
}
```

**Drop the redundant JPEG pass.** Step 1 is an in-process Tauri invoke on a 144×144 image that is re-encoded as JPEG for the device anyway, so the first lossy pass buys nothing:

```diff
-	if (active && slotContext) setTimeout(async () => await invoke("update_image", { context: slotContext, image: canvas.toDataURL("image/jpeg") }), 10);
+	if (active && slotContext) setTimeout(async () => await invoke("update_image", { context: slotContext, image: canvas.toDataURL() }), 10);
```

Minor and unrelated to the above: [`src-tauri/src/encoder_layouts.rs:42`](https://github.com/nekename/OpenDeck/blob/main/src-tauri/src/encoder_layouts.rs#L42) also uses `FilterType::Nearest`, on the fallback path taken when an action has no encoder config. The normal dial-strip path goes through `streamdeck-strip-render` at native 200×100 and is unaffected, which is why strips look noticeably cleaner than keys today.

Happy to open PRs for any of these.

constrictor:
Fixes #411.

Key images reach the panel with their antialiasing destroyed — stair-stepped diagonals, strokes whose width visibly alternates along their length, and small text reduced to mush. Found while writing an ALSA plugin that draws its key faces as SVG, on 2.14.0 with a Stream Deck +.

The cause is in `elgato-streamdeck`: `convert_image_with_format` resamples every key image to the device's key size with `FilterType::Nearest`. At 144 → 120, a 6:5 ratio, that drops every sixth row and column and point-samples soft edges instead of averaging them. I've opened OpenActionAPI/rust-elgato-streamdeck#62 for it.

This PR fixes it from OpenDeck's side so it doesn't have to wait on a crate release, and makes two related changes.

**`elgato.rs` — scale to the key size before handing the image over.** With the image already at the target size, the crate's `resize_exact` is a no-op in effect, so the filter that runs is ours. This stays correct once #62 lands: that PR skips the resample entirely when the image already matches, so the two compose rather than fight.

**`rendererHelper.ts` — send the canvas as PNG, not JPEG.** That hop is an in-process Tauri invoke on a 144×144 image which is re-encoded as JPEG for the device anyway, so the lossy pass bought nothing but artefacts.

**`encoder_layouts.rs` — same filter change** on the fallback image path. Low impact, since the normal dial-strip path goes through `streamdeck-strip-render` at native 200×100 and is unaffected, but it was the same `FilterType::Nearest`.

### Testing

Rendering the same source image through the full pipeline twice, changing only the filter and keeping both JPEG passes, the `Nearest` output matches what my hardware shows and the filtered output is clean, with no change to the artwork.

I could not build the tree here: `hidapi` needs `libudev` headers that aren't installed on this machine. I type-checked each changed expression against `image` 0.25.10 with the same imports the surrounding files use, and `rustfmt --check` is clean on both Rust files; the TypeScript line stays inside your 160-column Prettier width. So this wants a real build and an on-device look before it goes in — I'd expect the visible difference to be obvious on any device with a non-square-ratio key size.

### DIFF
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -61,7 +61,18 @@ pub async fn update_image(context: &crate::shared::Context, image: Option<&str>)
 				let (r, g, b) = extract_average_colour(&image::load_from_memory(&bytes)?);
 				device.set_touchpoint_color(context.position - key_count, r, g, b).await?;
 			} else {
-				device.set_button_image(context.position, image::load_from_memory(&bytes)?).await?;
+				// Scale to the key size ourselves. elgato-streamdeck would otherwise do it
+				// with FilterType::Nearest, which discards the antialiasing of whatever the
+				// plugin drew: at the Stream Deck +'s 6:5 ratio (144 -> 120) that drops every
+				// sixth row and column and point-samples soft edges into stair-steps.
+				let img = image::load_from_memory(&bytes)?;
+				let (kw, kh) = kind.key_image_format().size;
+				let img = if img.dimensions() == (kw as u32, kh as u32) {
+					img
+				} else {
+					img.resize_exact(kw as u32, kh as u32, image::imageops::FilterType::Lanczos3)
+				};
+				device.set_button_image(context.position, img).await?;
 			}
 		} else if context.controller == "Encoder" {
 			let mut img = image::DynamicImage::new_rgb8(200, 100);
--- a/src-tauri/src/encoder_layouts.rs
+++ b/src-tauri/src/encoder_layouts.rs
@@ -39,7 +39,7 @@ pub async fn generate_encoder_image(context: &crate::shared::Context, fallback: 
 			let mut fallback_canvas = RgbaImage::from_pixel(200, 100, Rgba([0, 0, 0, 255]));
 			let fallback_img = image::load_from_memory(fallback)
 				.context("Failed to decode fallback image")?
-				.resize(72, 72, image::imageops::FilterType::Nearest);
+				.resize(72, 72, image::imageops::FilterType::Lanczos3);
 
 			overlay(&mut fallback_canvas, &fallback_img.to_rgba8(), 64, 14);
 
--- a/src/lib/rendererHelper.ts
+++ b/src/lib/rendererHelper.ts
@@ -176,7 +176,8 @@ export async function renderImage(
 
 	context.restore();
 
-	if (active && slotContext) setTimeout(async () => await invoke("update_image", { context: slotContext, image: canvas.toDataURL("image/jpeg") }), 10);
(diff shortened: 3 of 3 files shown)


nekename:
> This PR fixes it from OpenDeck's side so it doesn't have to wait on a crate release

I maintain that library too, so there's no need to bring in a hotfix here as I can merge the PR upstream and then we can just bump the crate version here.

> rendererHelper.ts — send the canvas as PNG, not JPEG

What makes you think that this is a conversion step? It's unlikely that the browser engine is storing the canvas data as PNG. Just because the default export format is PNG doesn't mean that using JPEG here is an extra conversion step.

Let's do this upstream.

PR #419, opened 2026-08-10
fix: pre-resize key images with Lanczos3 to avoid Nearest-filter aliasing

danshapiro:
## Problem

The elgato-streamdeck crate's convert_image_with_format resizes with FilterType::Nearest for every device kind, so OpenDeck's 144x144 rendered keys get nearest-neighbor-downscaled to native key resolution — e.g. 80x80 on a Stream Deck Mini, a non-integer 1.8x ratio — producing visibly aliased/jagged icons.

## Fix

Pre-resize in elgato.rs to kind().key_image_format().size with Lanczos3 so the crate's resize is a no-op; also upgraded the encoder LCD fallback 72x72 resize from Nearest to Lanczos3 for consistency. Note the Infobar path already uses Lanczos3, so this aligns the remaining paths.

## Testing

Built and running on Windows against a Stream Deck Mini, 80x80 BMP keys — clearly visible smoothing improvement on real hardware. Author runs this patch daily.

## Alternative considered

Fixing the filter upstream in the elgato-streamdeck crate itself, but pre-sizing in OpenDeck works regardless of the crate's behavior and keeps the change local.

### DIFF
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -1,5 +1,6 @@
-use crate::encoder_layouts::generate_encoder_image;
 use crate::events::inbound;
+use crate::shared::{ActionInstance, Encoder, config_dir};
+use crate::store::profiles::{acquire_locks, get_slot};
 
 use std::collections::HashMap;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


danshapiro:
The CLA is a little too broad, so I'll just submit it as an issue in case
that's helpful. Thanks for the great code.

My Book <http://hotseatbook.com>   |   My Twitter
<http://www.twitter.com/danshapiro>   |   My Turtles
<http://www.robotturtles.com/>   |   My, it'd be nice to work together
<http://glowforge.com/jobs>

nekename:
An issue and PR have already been opened here as well as an issue and PR upstream on the library, so please follow those discussions instead of creating even more :sweat:

PR #429, opened 2026-08-19
Let a device place its encoders among the keys

tommasobbianchi:
## The problem

`DeviceView` always draws encoders in a row below the keypad. That is right for every Elgato
device and for the Stream Deck+, but not for hardware that puts a dial *in* the same strip as
the keys.

The VSD Stream Dock N1 is portrait: a strip along the top edge holds two small screens with
buttons behind them and a third with the knob under it, and fifteen keys sit below. Whichever
way the strip is mapped into the grid, the knob ends up in a row of its own beneath fifteen
keys it is physically above, so the deck on screen is laid out differently from the one under
the user's hands. It is not cosmetic in use: you reach for the dial by looking at the app and
find the wrong end of the device.

## The change

A device may now report `encoder_slots`: the keypad position of each encoder that is
physically among the keys. `encoder_slots[i]` is where encoder `i` is drawn.

* Listed encoders are drawn in place of that keypad cell.
* Anything not listed keeps the row below the keypad, exactly as now.
* The field defaults to empty via `serde_inline_default`, so **every existing device and every
  plugin that does not send it is unaffected** -- including the Elgato path, which passes an
  empty vector explicitly.

Grid navigation follows the same split: the encoder row is only counted as a row when
something is left in it, and `overflowsY` likewise.

## Testing

Built and run against a VSD Stream Dock N1 through a mirajazz-based plugin
([tommasobbianchi/opendeck-vsd-n1](https://github.com/tommasobbianchi/opendeck-vsd-n1)) that
registers `encoder_slots: [2]`: the dial is drawn in the top strip next to the two buttons, and
the fifteen keys fill the rows below, matching the hardware. An Elgato Stream Deck was not
available to me, so the unchanged path is argued from the code rather than measured -- worth a
second pair of eyes on that.

Happy to adjust the shape of the field. An alternative would have been a single "encoders come
first" flag, but that still leaves the dial on a row of its own rather than in the strip where
it belongs, so it would not have solved this device.

### DIFF
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -166,6 +166,8 @@ async fn init(device: AsyncStreamDeck, device_id: String) {
 				encoders: kind.encoder_count(),
 				touchpoints: kind.touchpoint_count(),
 				infobars: if kind == Kind::Neo { 1 } else { 0 },
+				// Every Elgato device keeps its dials in a row of their own.
+				encoder_slots: Vec::new(),
 				r#type: device_type,
 			},
 		},
--- a/src-tauri/src/shared.rs
+++ b/src-tauri/src/shared.rs
@@ -42,6 +42,11 @@ pub struct DeviceInfo {
 	pub touchpoints: u8,
 	#[serde_inline_default(0)]
 	pub infobars: u8,
+	/// Keypad positions that are physically a dial rather than a key, for devices where the
+	/// encoders sit among the keys instead of in a row of their own. Encoders listed here are
+	/// drawn in place; any not listed keep the row below the keypad.
+	#[serde_inline_default(Vec::new())]
+	pub encoder_slots: Vec<u16>,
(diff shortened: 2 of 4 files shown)


Terrorwolf01:
We had a similar PR just recently rejected (#427) in Favor for #359. So this is probably gonna be rejected also.

tommasobbianchi:
Closing this in favour of #359.

I opened this before reading the discussion on #427, where the same objection applies: adding a repositioning field per device quirk keeps the layout model rigid and gets messy fast. An explicit layout is the right level to solve it at.

I've posted the device this came from over on #359 as a concrete test case — it has a rotary encoder that sits *inline with two keys* in a strip above the keypad, and whose screen is a discrete 64x64 panel rather than a shared touch strip, which is a case the current draft doesn't quite cover yet.

Thanks @Terrorwolf01 for the pointer.

PR #430, opened 2026-08-19
Keep a key's state update in the copy that survives a re-render

tommasobbianchi:
Fixes the frontend half of #152, using the diagnosis @atomskz posted in that thread.

`Key.svelte` keeps two copies of a slot: the component-local `slot`, and the parent's
`profile.keys[]`, bound in as `inslot`. The `update_state` listener only wrote to the local one:

```ts
listen("update_state", ({ payload }) => {
    if (payload.context == slot?.context) slot = payload.contents;
});
```

So `profile.keys[]` kept the instance from `create_instance`, still carrying the action's
default image. The next time anything set `profile = profile` in `DeviceView` — after
`create_instance`, after `move_instance` — the reactive `update(inslot)` overwrote the fresh
`slot` with the stale one, and `Key` re-rendered and pushed the default image back to the
device.

This writes the update into both copies, and moves `lastInslot` with them so the write back
does not bounce straight through `update()` on the next tick.

## Testing

On a VSD Stream Dock N1, driving `setImage` for keys owned by the Starter Pack and Linux App
Launcher plugins (through `--process-message`, so the ownership check is not in the way).
Before: the image appeared and was replaced by the action's default icon on the next grid
re-render. After: it stays, survives a restart, and OpenDeck rasterises it to `0.png` as it
does for an image chosen in the editor.

I have not reproduced the original report's exact path (Stream Deck MK.2, disconnect and
reconnect, Multi OBS Controller) as I have none of that hardware, so this may not be the only
route to the symptom -- the reconnect path is worth a second look. It is one concrete cause,
and the one I could measure.

### DIFF
--- a/src/components/Key.svelte
+++ b/src/components/Key.svelte
@@ -51,7 +51,15 @@
 	}
 
 	listen("update_state", ({ payload }: { payload: { context: string; contents: ActionInstance | null } }) => {
-		if (payload.context == slot?.context) slot = payload.contents;
+		if (payload.context != slot?.context) return;
+		slot = payload.contents;
+		// The parent's `profile.keys[]` is the copy that survives a grid re-render, so a state
+		// update that only lands in the local `slot` is undone the next time anything sets
+		// `profile = profile` -- which is what made plugin-set images revert to the action's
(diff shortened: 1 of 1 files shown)


tommasobbianchi:
signed cla

PR #441, opened 2026-08-30
Turn touch strip swipes into dial ticks on the Stream Deck + and + XL

emaspa:
Swipes on the Plus touch strip are currently discarded (the `_ => Ok(())` arm of the device event loop). This turns them into dial rotation: a swipe becomes encoder ticks for the dial it started over, one tick per 12 px of horizontal travel. Dragging across a dial's panel moves its value by the drag distance, and any plugin that already handles `dialRotate` gets this for free; plugins see ordinary tick events, so there is no API or protocol change.

One limitation worth stating: the hardware reports a swipe once, at finger lift, with start and end coordinates, so the adjustment lands when the finger lifts rather than tracking the drag live. That is a firmware constraint, not a choice in this patch.

On the divisor: 12 px per tick makes a full 200 px panel drag about 16 ticks, which felt right against the physical dial's detents on my Stream Deck +, where I have been running this against real plugins. Happy to tune it, or gate the whole thing behind a setting, if you would prefer.

Vertical position is ignored on purpose; a swipe that crosses tile borders is attributed to the dial it started over.

### Preflight checklist

### DIFF
--- a/src-tauri/src/elgato.rs
+++ b/src-tauri/src/elgato.rs
@@ -220,6 +220,17 @@ async fn init(device: AsyncStreamDeck, device_id: String) {
 					};
 					inbound::devices::touchscreen_press(touchscreen_press(position, x, y, true)).await
 				}
+				DeviceStateUpdate::TouchScreenSwipe((sx, _sy), (ex, _ey)) => {
+					let position = match kind {
+						Kind::Plus | Kind::PlusXl => (sx / 200) as u8,
+						_ => continue,
+					};
+					let ticks = ((ex as i32 - sx as i32) / 12).clamp(-127, 127) as i8;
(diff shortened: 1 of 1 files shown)


Terrorwolf01:
Wouldn't this cause problems later on when the Profile rework is done, the Pages are implemented and then Swipe is used as on the official app to change Pages in a Profile?

emaspa:
> Wouldn't this cause problems later on when the Profile rework is done, the Pages are implemented and then Swipe is used as on the official app to change Pages in a Profile?

Possibly, I am not aware of that work.

nekename:
This is completely inconsistent with Elgato's behaviour and will create nothing but confusion.

